PR #63, opened 2015-01-02
taking query into account when calculating numDocuments and totalOccuren...

jamesdelonay:
...ces for correct reporting in case where query is used.

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -152,6 +152,7 @@ function serializeDoc(doc, maxDepth){
 
 var interimResults = {}; //hold results here until converted to final format
 // main cursor
+var numDocuments = 0;
 db[collection].find(query).sort(sort).limit(limit).forEach(function(obj) {
 	//printjson(obj)
 	flattened = serializeDoc(obj, maxDepth);
@@ -174,8 +175,8 @@ db[collection].find(query).sort(sort).limit(limit).forEach(function(obj) {
 			interimResults[key]['totalOccurrences']++;
 		}
 	}
+    numDocuments++;
 });
-
 
 var varietyResults = [];
 //now convert the interimResults into the proper format
@@ -185,11 +186,9 @@ for(var key in interimResults){
 	newEntry['_id'] = {'key':key};
 	newEntry['value'] = {'types':Object.keys(entry['types'])};
 	newEntry['totalOccurrences'] = entry['totalOccurrences'];
-	newEntry['percentContaining'] = entry['totalOccurrences']*100/limit;
+        newEntry['percentContaining'] = entry['totalOccurrences']*100/limit;
 	varietyResults.push(newEntry);
 }
-
-var numDocuments = db[collection].count(query);
 
 // We throw away keys which end in an array index, since they are not useful
 // for our analysis. (We still keep the key of their parent array, though.) -JC
@@ -207,9 +206,7 @@ var map = function(item) {
   }
   // we don't need to set it if limit isn't being used. (it's set above.)
   if(limit < numDocuments) {
-    var existsQuery = query;
-    existsQuery[keyName] = {$exists: true};
-    item.totalOccurrences = db[collection].count(existsQuery);
+      item.totalOccurrences = db[collection].count(query);
   }
   item.percentContaining = (item.totalOccurrences / numDocuments) * 100.0;
   return item;
--- a/variety.js
+++ b/variety.js
@@ -189,7 +189,7 @@ for(var key in interimResults){
 	varietyResults.push(newEntry);
 }
 
-var numDocuments = db[collection].count();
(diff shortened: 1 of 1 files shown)


jamesdelonay:
Great tool. I was using it to look at variety of various queries when I noticed the issue.

todvora:
Hi @jamesdelonay,
Thank you for your interest in Variety. Documentation about the query param is little bit ambiguous. Until now I always thought about the query as a possibility to filter analyzed documents, where the rest is calculated in the same way as by ```limit``` param. From limit README:

> the tool calculates totalOccurrences and percentContaining from all the collection's documents, even those outside the "limit".

So Variety now handles the query in the same manner. Filters analyzed and does not influence total numbers. But maybe the current behavior does not make any sense. It probably depends on use cases. 

What do you think @JamesCropcho and @wfreeman? We should clearly define, what is the expected behavior. 

Thank you,
Tomas

jamesdelonay:
Hi Tomas,

I was assuming that the expected behavior would be to reflect the variety
of the query since that is really all that is being looked at. Reporting
the variety compared to the total number of records in collection may give
users a false sense of its composition.

Thanks,
James Delonay

JamesCropcho:
Tomáš and James,

I agree that there is ambiguity, and I believe it can be pretty easily resolved by either changing the name of "query" to something more precise, OR to do the above and also add a second precise parameter for James' proposed functionality, OR to remove the current "query" functionality and replace it with James', but with a more precise parameter name. Hopefully that makes sense.

Another option would be on parameter, with a switch to determine which query type.

Of course, naming and naming changes are to be taken very seriously (though I don't mind breaking anything to improve future UI).

Thoughts, including what they might be named?

Let Me Know, and Thanks Sincerely,
James

jamesdelonay:
Hello James and Tomáš,

I have recently pushed another change to my branch. It now takes the
effects of limit into consideration in cases with and without a query. I
hope that you will consider pulling this into variety.js.

Let my explain why I think that it would be beneficial. First, the output
is not even correct in either the case where queries or limits are used.
Often times the total occurrence count of a field will either be zero or
greater then the number of records that are considered. This is clearly a
flaw. The second, it is inaccurate in its reporting of the field scores.
Just try a limit of one and you will see %100 of all documents in the
database are reported to have the fields that are present in this one
record examined. It would be far safer to report the number field
occurrences with respect to the actual number documents investigated.
Nothing needs to be changed, because in its present state it is not even
valid.

Your variety tool for mongodb has been so useful to me and I hope that you
will allow me to give something back.
Thanks for your consideration,
-James Delonay

JamesCropcho:
Hello James D.,

Thank you sincerely for your hard work on this.

Apologies for the delay in our reply. It's been a busy week and will get back to you as soon as we're able.


-James C.

jamesdelonay:
Hey James,

No problem. If you want me to further clarify what I am trying to
accomplish, let me know. I am often times not the best communicator, so I
understand if you want more explanation.

Thanks,
James D.

JamesCropcho:
@todvora @wfreeman I feel like this is a non-trivial change. Do either of you have any thoughts here?

todvora:
Hi,
I had exactly the same feeling as @jamesdelonay, when I was writing the tests. Both ```limit``` and ```query``` are currently extrapolation. But the plain sum of results in selected range seems more natural and expected. 

When someone wants to analyze really only one part of the collection, there is currently no possibility, because the results are always extrapolated to all documents. And it's easier to compute manually the extrapolation from real results, then the other way round.

I would vote for this change. 

Best regards,
Tomas

JamesCropcho:
Tomáš, if you feel that @jamesdelonay has the right idea, I am happy to put my support behind merging this pull request then, unless we hear something from Wes to the contrary.

Otherwise, to move forward- as we would hope(!) this pull causes three tests not to pass, and therefore our build to fail.

Tomáš, would you prefer to update the tests yourself or should we ask @jamesdelonay, perhaps? I'm open-minded- whichever you think is best.

-James

todvora:
No problem, I can update the tests. Of course if @jamesdelonay wants to do that, its perfectly fine too. 
T.

JamesCropcho:
I fine either way, of course. Whatever is best for both of you.

jamesdelonay:
I vote yes. ;)

todvora:
Hi @jamesdelonay,
Sorry that it took us so much time, but I finally managed to merge you pull request and adapt the test cases. Thank you very much for your contribution!

Now we have to update outdated readme, incorporate behavior changes and replace output examples with the ascii results. Working on it ;)

Have a nice day,
Tomas

JamesCropcho:
I agree with Tomáš- thank you very much for your contribution, James. It is a significant improvement for Variety. I hope that you and many others benefit from it.

PR #64, opened 2015-01-06
Added Object.keys function if the native implementation lacks of it

### LINKED ISSUE #62: Object.keys is not a function variety.js;186
Hi All,

I'm trying to run variety.js on a database collection and getting the following error.

Mon Dec 29 12:50:05 TypeError: Object.keys is not a function variety.js:186
failed to load: variety.js

It seems to be an issue when it's converting interim results.  The error producing line is here:

```
newEntry['value'] = {'types':Object.keys(entry['types'])};
```

I'd be happy to investigate this a bit more, but I'm not sure what the Object is referencing.  Any help would be appreciated.

Thanks,
Marc

malderete:
This patch allows the tool to work with old mongodb versions, in some cases a version upgrade is not an option so we could continue using this tool. This patch fix the closed #62 .

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -9,6 +9,50 @@ Please see https://github.com/variety/variety for details.
 
 Released by Maypop Inc, © 2012-2014, under the MIT License. */
 
+//To add compatible Object.keys support in older environments that do not natively support it (old mongo versions).
+// From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys
+if (!Object.keys) {
+  Object.keys = (function() {
+    'use strict';
+    var hasOwnProperty = Object.prototype.hasOwnProperty,
+        hasDontEnumBug = !({ toString: null }).propertyIsEnumerable('toString'),
+        dontEnums = [
+          'toString',
+          'toLocaleString',
+          'valueOf',
+          'hasOwnProperty',
+          'isPrototypeOf',
+          'propertyIsEnumerable',
+          'constructor'
+        ],
+        dontEnumsLength = dontEnums.length;
+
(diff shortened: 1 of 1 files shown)


JamesCropcho:
Hi Martin,

Thank you for patching Variety to allow it to work with older versions of MongoDB.

Well, I have used Martin's version locally, with my modern version of MongoDB and it's worked perfectly. @wfreeman and @todvora, do you see any issues with accepting the merge?


James

todvora:
Hi @malderete,
Thank you very much for your contribution and testing of Variety. I like the idea of running Variety also on older versions of MongoDB with a simple fix. Anyway I would recommend to use little bit different approach. 

Since we use ```Object.keys``` only once and on our own constructed object, we can avoid using it at all. It requires only minor change of source code without adding any new complex logic and functions. 

If you could find any time to review or test code in pull request https://github.com/variety/variety/pull/65, it would be great.

Best regards, Tomas

JamesCropcho:
Thanks for discovering this, Martin. Everybody who runs an older MongoDB version will be very thankful :)

malderete:
@JamesCropcho , @todvora  thanks for reviewing this PR, for the feedback and for the interest in this feature :).
My first patch was the esiest approach to run velocity in an old MongoDB, so as @todvora said we use only once the Object.keys call and we can avoid it at all.
@todvora ,Ok I will review the #65 and let you know.
Best regards, Martin

todvora:
Hi @malderete, I received no complaints and the tests are passing, so I merged #65. I hope it works well for you. Please don't hesitate to contact me, if you have any thoughts on this.

Thank you,
Tomas

PR #65, opened 2015-01-12
Removed Object.keys call

todvora:
Object.keys call prevented to use Variety on older versions of MongoDB.


JamesCropcho:
I'm certainly in favor of this- seems like all upside and no downside to me.

todvora:
No complaints, tests are passing, so I am going ahead and merge this minor change.

PR #71, opened 2015-02-18
Try to reformat the patch

oz123:
I think this is better.

I does not include the README. Or the nice output in the commit message. But you can copy them manually.

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -68,12 +68,21 @@ if(typeof query !== 'undefined') {
 }
 log('Using query of ' + tojson($query));
 
-var $limit = db[collection].find($query).count();
-if(typeof limit !== 'undefined') {
-  $limit = limit;
-  limit = '_undefined';
+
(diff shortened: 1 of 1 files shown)


JamesCropcho:
Hello Oz ( @oz123), 

Thanks much for trying to reformat the pull request. However, thanks to @todvora's refactoring, which looks very good, I sadly don't think this patch works any more, in its current state. Very sorry about that, Oz.

However, @todvora did point us all in the right direction regarding how one would implement this feature post-refactoring: https://github.com/variety/variety/pull/72#issuecomment-75236692.

If you found time to re-do the patch, of course we'd be very happy, however given the circumstances I'd also totally understand if it exceeds your threshhold for time/energy on a FOSS project, Oz. No worries there if so.

Regardless, I thank you for all the you've provided, @oz123, in terms of the bugfix and the documentation issue and all the rest. --Regards, James

Issue #69, opened 2015-02-18
Critical Bug: Wrong percentage indications?


oz123:
Sometime variety would out put stuff like this:
````
[66] oti@core-dev:~/Software/variety  [master]  $ /srv/mongodb/bin/mongo  --eval "var db_name = 'client'; var collection = 'rules'" variety.js
MongoDB shell version: 2.6.4
connecting to: test
Variety: A MongoDB Schema Analyzer
Version 1.4.1, released 14 Oct 2014
Using query of { }
Using limit of 84
Using maxDepth of 99
Using sort of { "_id" : -1 }
Using outputFormat of ascii
Using persistResults of false
removing leaf arrays in results collection, and getting percentages
+----------------------------------------------------------------------------------------------+
| key                                      | types          | occurrences | percents           |
| ---------------------------------------- | -------------- | ----------- | ------------------ |
| rules.XX.filter                          | Object         | 5907        | 7032.142857142857  |
| rules.XX.filter.account                  | String         | 5907        | 7032.142857142857  |
| rules.XX.modify                          | Object         | 5907        | 7032.142857142857  |
| rules.XX.name                            | String         | 5907        | 7032.142857142857  |
| rules.XX.modify.channel                  | String         | 5012        | 5966.666666666666  |
| rules.XX.modify.sub_channel              | String         | 5012        | 5966.666666666666  |
| rules.XX.modify.sub2_channel             | String         | 4976        | 5923.809523809524  |
| rules.XX.modify.top_channel              | String         | 4592        | 5466.666666666666  |
| rules.XX.filter.channel                  | String         | 4278        | 5092.857142857143  |
| rules.XX.filter.campaign                 | String         | 3816        | 4542.857142857143  |
| rules.XX.modify.paid_performance_channel | Boolean,String | 1987        | 2365.4761904761904 |
| rules.XX.modify.exclude                  | Boolean        | 1164        | 1385.7142857142858 |
| rules.XX.modify.sea_brand                | String,Boolean | 567         | 675                |
| rules.XX.modify.cpc                      | Number         | 476         | 566.6666666666667  |
| rules.XX.modify.cost                     | Number         | 281         | 334.5238095238095  |
| rules.XX.modify.campaignId               | String         | 168         | 200                |
| rules.XX.modify.crr                      | Number         | 159         | 189.28571428571428 |
| _id                                      | ObjectId       | 84          | 100                |
| author                                   | String         | 84          | 100                |
| rules                                    | Array          | 84          | 100                |
[...]

PR #72, opened 2015-02-19
#69 - fixed wrong percentage indications in unnamed objects inside arrays

todvora:
The analysis of documents now runs in two steps. 

First there are aggregated keys and types of one document. If there are two or more same keys (due to unnamed objects inside arrays, containing .XX.), they are merged into one item. This prevents counting the key multiple times for one document.

Complete results are then aggregation of this pre-merged document fingerprints. 

Also, the array of types is now alphabetically ordered to remove randomness (and failing tests).

### DIFF
--- a/test/src/test/java/com/github/variety/test/SampleData.java
+++ b/test/src/test/java/com/github/variety/test/SampleData.java
@@ -23,7 +23,7 @@ class SampleData {
             "| name               | String       | 5           | 100      |\n" +
             "| bio                | String       | 3           | 60       |\n" +
             "| birthday           | String       | 2           | 40       |\n" +
-            "| pets               | String,Array | 2           | 40       |\n" +
+            "| pets               | Array,String | 2           | 40       |\n" +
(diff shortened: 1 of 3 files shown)


todvora:
I also refactored a little bit the computation process. The interimResults is removed and everything run in one pipe of map/reduce/filter functions. I hope, that it is more readable. Plus, the 'Random sampling of large collections' can be now very easily integrated (not done in this PR). All, what is needed is to add one more filter to pipe. Something like that could work: 

```
.filter(function(){return Math.random() < 0.1})
```
This should analyze only 1/10 of all documents in collection. All others will be skipped without processing. 

What do you think? Do you found it readable / maintainable?

Thank you,
Tomas

JamesCropcho:
Hello Tomáš, I quite like the idea of your unified pipe/filter setup. Very cool, and yes, will make further improvements even easier :)

I'm merging into `master` and closing.  --James

PR #75, opened 2015-02-28
Performance fix

todvora:
Hi,
I found out, that the processing pipe is complete performance killer. For small collections is it good enough. But analysis of collection with 500 000 results and enough keys never finishes. I compared results of analysis based on this [source collection](http://mongodb-enron-email.s3-website-us-east-1.amazonaws.com/).

Before the [wrong percentage indications in unnamed objects](https://github.com/variety/variety/commit/dd9b069a8aa6c36dd44208eac4276907725299b2) took Variety +- 50s to finish. After fix was it almost 2x more. And the pipe refactoring killed it totally. 

I didn't realized, that JS behaves differently than Java 8 Streams. In Java, the pipe/stream is processed lazilly. In JS is first computed the whole array of results and then send to the next method in the row. JS then built giant arrays and took to much memory and time. 

So I had to rewrite processing one more time, excluding those giant arrays. It lost little bit of elegance, but still passing all tests and performance is the same or better than before unnamed objects fix. The sampling could be still very easily in the reduce method. 

I am sorry to bother you once more time, but could you look at the code and review it?

Thank you,
Tomas

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -175,24 +175,63 @@ var serializeDoc = function(doc, maxDepth) {
   return result;
 };
 
-var mergeArrays = function(a, b) {
-  if(typeof a === 'undefined') {a = [];}
-  return a.concat(b) // merge two arrays into one, including duplications
-        .filter(function(item, pos, self){return self.indexOf(item) == pos;}) // remove duplications
-        .sort(); // sort alphabetically
-};
-
 // convert document to key-value map, where value is always an array with types as plain strings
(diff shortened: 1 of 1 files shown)


JamesCropcho:
Hi Tomáš, sorry to hear the pipe/stream stuff was not performant. But to learn how quickly we learned the mistake makes me feel happy and confident about Variety's future. I reviewed the pull request, and everything worked and made sense, so I merged.

Issue #76, opened 2015-03-04
Using limit produces wrong percentage


Acetolyne:
When you use limit = * the percentage seems to be based on the number you enter for the limit even when the results yield less results than the limit number. I believe percentage should not be based on the number you are looking for but rather the number which is returned.

JamesCropcho:
Hello Jon,

Thanks for using Variety, and for getting in touch.

If the percentage returned were relative to the entire collection's size, it would have to be listed with some caveat, such as "Percentage (estimated)", because it really is an estimate- a prediction- rather than definitive in the way the percentage as given is currently definitive.

I could see the neccesary use of "estimated" or equivalent to be confusing in its own right.

For the above, I am inclined to think that the current way is superior. However, I remain open-minded to further thoughts you have, as well as seeing other people "plus-one" this thread, indicating additional interest.

Thank You,
James

todvora:
Hi @Acetolyne, @JamesCropcho,
The issue is - when have 5 documents in your collection and limit set to 10, then in the results is correct occurence of 5, but incorrect percentage 50. The percentage computation is based on supplied number instead of real count of documents. 

I hope I fixed it in https://github.com/variety/variety/pull/79.

@Acetolyne: could you please verify, that the version from PR is working as expected?

Thank you both!
Tomas

Acetolyne:
@todvora Thanks I think I did not communicate the problem well #79 seems to fix the problem

PR #79, opened 2015-03-06
Fix documents percentages, if the limit is higher than real count of the documents #76 

todvora:
Fixes issue mentioned in https://github.com/variety/variety/issues/76

### DIFF
--- a/test/src/test/java/com/github/variety/test/LimitResultsAnalysisTest.java
+++ b/test/src/test/java/com/github/variety/test/LimitResultsAnalysisTest.java
@@ -35,4 +35,20 @@ public class LimitResultsAnalysisTest {
         analysis.validate("name", 1, 100, "String");
         analysis.validate("someBinData", 1, 100, "BinData-old");
     }
+
+    @Test
+    public void verifyLimitOverMaxResults() throws Exception {
+
+        // limit is set to higher number, that the actual number of documents in collection
+        // analysis should compute percentages based on the real number of documents, not on the
+        // number provided in the limit var.
+        final ResultsValidator analysis = variety.withLimit(10).runDatabaseAnalysis();
+
+        analysis.validate("_id", 5, 100, "ObjectId");
+        analysis.validate("name", 5, 100, "String");
+        analysis.validate("bio", 3, 60, "String");
+        analysis.validate("pets", 2, 40, "String", "Array");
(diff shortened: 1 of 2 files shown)


JamesCropcho:
@todvora and @Acetolyne, as I tried to explain in https://github.com/variety/variety/issues/76#issuecomment-77456020, I think this new output would be misleading. Maybe I didn't do a great job of explaining.

At best, this approach provides an estimate of the percentage occurrence in the collection as a whole, with wildly varying confidence based on the limit/collection-size ratio. And yet we are stating factually that we're displaying the precise percentage of the collection if we merge this pull request.

Does the above make sense?

If you feel that the percentages as displayed on `master` are confusing I'm open to alternatives such as listing "percentage (estimated)," perhaps. However, percentage as currently displayed on `master` is correct, relative to the query limit.

Thoughts?

todvora:
Hi James, i think we do talk about two different things. The problem here
is following : you have a collection that contains only 5 documents. In the
Variety params then specify, that limit is 10. Our results are then saying,
that _id contains only 50 procent of the documents. But there are really
100%, 5 of them analysed, all with the id field. This happens only when the
limit is set to higher number than the real count of documents in the
result set (after query filtering).The fix corrects it, counting the real
returned documents instead of simply applying the passed limit number to
percentage computation. There is no gues, no estimation, just the real
count of analysed results.

Current limit number was not validated anyhow and it could lead to nonsense
percentages, when you passed too high number. The percents should be the
same on the collection of 5 documents, when you pass the limit of 10, 15 or
100. There are still only 5 documents to analyse, but we returned each time
different result - 50%, 33%, 5%.

When the limit is set lower than max count of documents, nothing is
changed. The results are displaying percentages based only on the returned
results, ignoring the total count of documents in the collection, ignoring
the rest of documents, that were filltered out. No estimation based on
total count.

Does it make sense?

Thank you, Tomas
Dne 6.3.2015 22:03 "James Cropcho" <notifications@github.com> napsal(a):

JamesCropcho:
Sincere thanks for taking the time to explain, Tomáš.

I agree than `when limit > collection_size` we should `set limit = collection_size`. If this pull does that, very good! And if ,"[w]hen the limit is set lower than max count of documents, nothing is changed," very good, again!

I am happy to merge this pull request and I think it's a great addition. Thank again to Jon, too.

PR #91, opened 2015-04-28
Added latex option to outputFormat

joggino:
Added the latex option, useful for publications, also added ability to select number of columns for which you can select from left to right, i.e. you only want the first 2 columns, insert option numColumns = 2

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -19,6 +19,7 @@
     "maxDepth": true,
     "sort": true,
     "outputFormat": true,
+    "numColumns" : true,
     "persistResults": true
   },
   "latedef": true,
--- a/variety.js
+++ b/variety.js
@@ -317,8 +317,8 @@ if($outputFormat === 'json') {
      * Cut down the columns by returning space, or what should be returned
      * @type {string}
      */
-    output += row.map(function(cell, i) {return (i >= $numColumns ? '' : (i === 0 && ri === 1 ? '\\' + cell : cell) +
-                                                                 (i >= ($numColumns - 1) ? ' ' : ' & '));}).join(' ');
+    output += row.map(function(cell, i) {return i >= $numColumns ? '' : i === 0 && ri === 1 ? '\\' + cell : cell +
+                                                                 i >= $numColumns - 1 ? ' ' : ' & ';}).join(' ');
     output +=' \\\\ \\hline \n';
   });
 
--- a/variety.js
+++ b/variety.js
@@ -279,18 +279,19 @@ if($persistResults) {
   resultsDB[resultsCollectionName].insert(varietyResults);
 }
 
+  var table, maxDigits, output, significantDigits;
 if($outputFormat === 'json') {
   printjson(varietyResults); // valid formatted json output, compressed variant is printjsononeline()
 
 }else if($outputFormat === 'latex') {  // output nice latex table with results
-  var table = [['key', 'types', 'occurrences', 'percents']]; // header row
+  table = [['key', 'types', 'occurrences', 'percents']]; // header row
   // return the number of decimal places or 1, if the number is int (1.23=>2, 100=>1, 0.1415=>4)
-  var significantDigits = function (value) {
+  significantDigits = function (value) {
     var res = value.toString().match(/^[0-9]+\.([0-9]+)$/);
     return res !== null ? res[1].length : 1;
   };
 
-  var maxDigits = Math.max.apply(null, varietyResults.map(function (value) {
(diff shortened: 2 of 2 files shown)


todvora:
Hi @joggino,
Thank you for your contribution and for the efforts to make it passing jshint :-) I like the idea of another output format. I have only one objection to the ```numColumns``` param. It seems to me confusing (not clear, that it belongs to the latex output) and too specialized. 

@JamesCropcho, @wfreeman:
what do you think about the latex extension? Would it be useful also for other users? It seems, that latex is a good fit with data analysis, one of typical  use-cases of Variety. 

If we decide to accept this PR,  I'll be glad to little bit polish the source code, add tests and update the readme. 

Thanks,
Tomas

JamesCropcho:
Thanks for helping make Variety better, Joe.

Tomáš, I think that LaTeX support would be a great thing and I support adding this feature, particularly because of your generous offer to tighten/polish if need be.

I too believe that `numColumns` could be named better at the least, because I do not have an idea what it should do from its name ;)

Any counter-proposals for the name of this parameter?

joggino:
Hi James/Tomas,

I'd happily implement `numColumns` with a different name and for all outputs if necessary.

It's purpose is so that the user can select how many columns from left to right they wish to include, as in my case I only really needed the first 2.
It just seemed easier to automate than having to remove code once its generated in the terminal.

Another input I'd like to try and add would be the ability to output a Microsoft word table, through the use of open xml (http://officeopenxml.com/WPtableGrid.php). I'd need to research it more but it could help in table creation if people prefer to use a word editor.

JamesCropcho:
Hello Joe, since you seem in favor (at least for now; not signing you up for anything against your will!) with the plugin model, I'm going to close this for now. But feel free to re-open at will if the plugin thing doesn't happen. —James

joggino:
I'm now using the plugin system @todvora has created, it's in keeping with lose coupling and doesn't bloat the initial files. 
I'm also creating a node command line script which executes variety and plugins as a child, then saves the output to a file. It's currently very rudimental, however once I have got my thesis handed in I will round off the node script so anyone can save output to a file.

JamesCropcho:
Great to hear you're liking @todvora's plugin so far. Let us know any feedback kindly. And thanks for your work on it!

Regarding the commandline script, you may want to check out variety-cli (https://github.com/variety/variety-cli) if you have not yet done so.

Yes, I figured that anyone wanting to save output might pipe it from STDOUT to a file, but alternatives could be good, too.

You may wish to run potential enhancements past the Variety community (including @todvora and me) prior to doing a lot of work constructing them, in case anyone sees an issue with the suggested functionality, or it already exists in the Variety ecosystem.


Thanks Again,
James

joggino:
I didn't see that CLI before you posted it, with regards to the STDOUT that's all I'm doing, but with the use of plugins, which the CLI doesn't currently support. 
Is there currently a function to print out all collection schemas? I was thinking of building that into the node app, this is mainly just so I don't have to manually draw my tables for my final thesis.

Acetolyne:
There is not at this time a way to print out all collections, I at one point had code to where it would take an array of collections as a parameter but the guys did not know if it was proper JSON, I simply looped there output for each collection name besides the Variety header. You can see what I did on my fork on the featured branch. Perhaps you can do a better job than I did and get this implemented, I also had plans to allow it to run on all collections minus the system collection unless the user specified that they want it included. I hope you can get this implemented in a way they will accept since it would be a great feature for Variety to have. Good luck on coding something they will accept.

todvora:
Hi @Acetolyne,
I had your requests in my mind, when I thought about [the plugin infrastructure](https://github.com/todvora/variety/tree/plugin-infrastructure). I believe we will be able to give you some hooks on the config and on the [main cursor](https://github.com/variety/variety/blob/master/variety.js#L257) to allow you to cycle through all collections. Plugins will have separated tests infrastructure so they don't have to deal with the complicated Maven test suite. 

I hope you will be interested in creation of analyse-all plugin then.

Thank you,
Tomas

JamesCropcho:
Hi all,

Yes, I think the plugin support will be very useful for features like what @joggino and @Acetolyne are seeking.

It's true that the CLI doesn't support plugins yet—the plugins paradigm is very new for Variety so it can't bubble everywhere it should all at once, sadly, but it's exciting that we're in the middle of a great step forward for all of us.


—James

Acetolyne:
I do agree the plugin idea was a good one. Almost all my projects have plugin capabilities from the beggining it will make the project able to scale much easier. Thx for your hard work on the plugin ability.your time in making it is very much appriciated.

Issue #92, opened 2015-04-29
Plugin infrastructure


todvora:
Hi all,
Recently we received a PR for the latex support. Requests like this requires lots of coding from the author, long discussions under the PR, dealing with test suite of Variety and is complicated on all sides. Maybe we should come up with some form of a plugins infrastructure, that would allow contributors extending Variety functionality without modifying the main library itself. 

I have tried to integrate a plugins functionality and it could be done with only minor changes of Variety. You can see the first implementation [here](https://github.com/todvora/variety/tree/plugin-infrastructure). 

I also created one [sample plugin](https://github.com/todvora/variety-plugin-csv), that outputs results in the CSV format and has configurable delimiter. In the same way could be implemented also [the latex output](https://github.com/todvora/variety-plugin-latex).

Plugins can be attached simply by adding a parameter like this:
```
mongo test --eval "var collection = 'users'; var plugins='/path/to/plugin/index.js';" variety.js
```
More details can be found in the sample plugin readme. The dynamic loading of plugins is based on the ```load``` function provided by the Mongo shell.  

The sample plugin is compatible with the [NPM](https://www.npmjs.com/) specification, but does not depend on node/npm anyhow. It allows easier testing and providing metadata in a standard format. Around this format can be later build a variety-plugin directory (for example like [gitbook plugins](http://plugins.gitbook.com/)).

The main Variety library is still standalone, no dependencies or additional requirements. Without the plugin param behaves exactly the same way as before. 

What do you think? Does it make sense? 

Thanks,
Tomas

joggino:
I'm in favour, it makes the process a lot simpler to use and people can add what they need without bloating the initial library for everyone.

JamesCropcho:
I think that sounds like a brilliant idea, Tomáš. And I do like that it is NPM-compatible and also(!) not requiring any additional runtime dependencies.

Yes, in theory we could "dog food" the plugin functionality, meaning we would take all but the core functionality out of Variety core and into plugins, themselves. But that's based on everyone time availability and is certainly not a dependency, just a thought.

I'm amenable to moving forward on this.


James

todvora:
Hi all,
just up-to-date info:

[...]

PR #96, opened 2015-05-11
Plugin infrastructure

todvora:
onConfig and outputFormat hooks for plugins, config read unified. See also #92.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -13,13 +13,16 @@
     "BinData": false,
     "DBQuery": false,
     "printjson": false,
+    "load": false,
+    "module": false,
 
     "query": true,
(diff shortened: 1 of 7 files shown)


JamesCropcho:
This is all very exciting, man! Don't you think it's long, long overdue for a pretty big version bump ;) @todvora?

todvora:
Thanks! You are right about the version number. It is now changed from 1.4.1 to 1.5.0. I added functionality in a backwards-compatible manner and thus the minor version should be incremented (due to [semver](http://semver.org/)).

JamesCropcho:
Happily merged, man! The start of a new epoch.

PR #101, opened 2015-07-30
add rounding percents

saromanov:
Hi!
I just added rounding of percentages in the output table. It would be useful if, we have many numbers after the dot in the `percents` column . Example:

```
+-------------------------------------------------------------------------------+
| key                        | types        | occurrences | percents            |
| -------------------------- | ------------ | ----------- | ------------------- |
| _id                        | ObjectId     |          11 | 100.000000000000000 |
| name                       | String       |          10 |  90.909090909090907 |
| bio                        | String       |           3 |  27.272727272727273 |
```

After applying 

```
mongo test --eval "var collection = 'users',round=true" variety.js
```

The table will look like:

```
--------------------------------------------------------------------+
| key                        | types        | occurrences | percents |
| -------------------------- | ------------ | ----------- | -------- |
| _id                        | ObjectId     |          11 |    100.0 |
| name                       | String       |          10 |     91.0 |
| bio                        | String       |           3 |     27.5 |
```

What do you think about this?

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -71,6 +71,7 @@ var readConfig = function(configProvider) {
   read('sort', {_id: -1});
   read('outputFormat', 'ascii');
   read('persistResults', false);
+  read('round', false);
   return config;
 };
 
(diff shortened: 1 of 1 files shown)


JamesCropcho:
Hello Sergey, thanks for the pull request!

I would assume it's not much harder to change the parameter `round` from `true` to accepting any integer greater than or equal to one. If so, thoughts on that change?

Also, I'm wondering if this would go in a plugin. Maybe, maybe not? @todvora , you have any thoughts on the matter?

Thanks again. We appreciate the help!


James

saromanov:
OK, but i think ```round=true``` is more clear, than numbers. We just make a better view :) At the end, this is not most important feature.

JamesCropcho:
Closing due to inactivity, but feel free to reopen, Sergey!

PR #111, opened 2016-03-15
Tests rewritten to JavaScript, npm package definition, ESLint

todvora:
This PR tries to simplify testing of variety and has some nice side effects.

- All the Java code is now gone. No more Java or Maven requirements for testing. They have been replaced by JavaScript (details lower). This speed things up, hopefully lowers bariers to run, modify and add new tests. It also means no troubles with Oracle (see [Running Java on Docker? You're Breaking the Law](http://blog.takipi.com/running-java-on-docker-youre-breaking-the-law/))
- The package.json has been introduced. This makes Variety project NPM + Node compatible. It is now possible to provide variety as a dependency (for example in variety-cli), can be distributed through NPM, standardizes development dependencies and process, provides all necessary scripts, makes releasing easier (even for GitHub). 
- JSHint has been replaced in favor of [ESLint](http://eslint.org/). ESLint is currently the superior tool for linting (and automatically fixing!) formatting and code style issues. It has also built-in support for MongoDB shell scripts. ESLint is now used to validate both variety.js and test scripts. [Read more about ESLint vs JSHint](http://www.sitepoint.com/comparison-javascript-linting-tools/#eslint). Execute by calling ```npm run lint``` or ```npm run lint:fix```
- Tests are written in pure JavaScript ([ES6 / ECMAScript 2015](https://github.com/lukehoban/es6features)), executed by [Mocha.js](https://mochajs.org/) with [Babel compiler](https://babeljs.io/docs/learn-es2015/) in the background. There is a very thin wrapper around Variety to handle command line input/output and some test helpers (init, cleanUp) in directory ```spec/utils```. 
- Tests are still fully executable locally and without Docker. In this case, running MongoDB instance + mongo shell + node/npm is required. Execute by calling ```npm test```.
- Dockerfile has been simplified. It's still based on official MongoDB images, where only Node.js is added and tests linked. The build is much faster and every other execution of tests is really fast. Execute by calling ```npm run test:docker```

The current setup also open doors for future improvements. We can very easily tag releases, make Travis-CI to test and deploy them on NPM or GH. We can split variety.js into small, readable and testable modules, use latest JS features and then let it compile to release version by [Browserify](http://browserify.org/). We can include plugins in Variety repository and distribute them along with Variety itself, bundled in one minimized file (it would be much easier to really write and use plugins). All non-core code can be moved to plugins (like ascii formatting)...

For the users itself, nothing changes. They still have the one variety.js file, same command line interface, still no dependencies at all. 

What do you think? Is it the way to go? Any requirements or drawbacks I missed? Let me know.


JamesCropcho:
Wow, cool, Tomáš. In a perfect world this pull request would actually be broken into several, but whatever!

On behalf of all the users of Variety, thanks much, man. I think these are all steps in the right direction for the tool's continued relevance and growth.

I'm accepting this PR. If anybody else has further thoughts feel free to chime in.

Happy March, all!

PR #114, opened 2016-04-01
ES6 and ES7 features in tests (import, export, async & await)

todvora:
Let's bring in some ES6 and ES7 goodies to make the tests even easier to read, understand and write. The promises boilerplate replaced by async / await and requires replaced by ES6 imports / exports.

What do you think? Step in the right direction or nonsense?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
- "presets": ["es2015" ]
+ "presets": ["es2015", "stage-0"]
 }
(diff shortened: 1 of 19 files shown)


JamesCropcho:
Tomáš, I'm pretty sure I know what the answer is, but: merging this PR would not change that that, to use Variety, one needs no other dependencies than MongoDB, right?

todvora:
No, those changes and dependencies are just for the tests. Variety is still dependencies-free and standalone.

JamesCropcho:
Great, as I figured.

Alright, I'll happily merge this PR. Is it a step in the right direction, or nonsense? Well, it's certainly not nonsense, and much of this is a matter of opinion. But if you're down for this change, I'm cool with that!

PR #140, opened 2018-01-26
Update year to 2018

JamesCropcho:


JamesCropcho:
…also includes a version bump which we somehow missed in October.

PR #142, opened 2018-02-03
New 'lastValue' property.

omachala:
Details are written in README.markdown

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -137,31 +137,24 @@ You can add ```lastValue``` property to show values of the last document.
 
     $ mongo test --eval "var collection = 'orders', lastValue = true" variety.js
     
-    +-----------------------------------------------------------------------------------------------------+
-    | key                                  | types    | occurrences | percents | lastValue                |
-    | ------------------------------------ | -------- | ----------- | -------- | ------------------------ |
-    | _id                                  | ObjectId |           1 |    100.0 | 5a5e7f64956ad87a681c13fb |
-    | my_date                              | Date     |           1 |    100.0 |            1516142436000 |
-    | order                                | Object   |           1 |    100.0 |                          |
-    | order.billing_address                | Object   |           1 |    100.0 |                          |
-    | order.billing_address.address1       | String   |           1 |    100.0 | 123 Fake Street          |
-    | order.billing_address.city           | String   |           1 |    100.0 | Fakecity                 |
-    | order.billing_address.country        | String   |           1 |    100.0 | Canada                   |
-    | order.customer                       | Object   |           1 |    100.0 |                          |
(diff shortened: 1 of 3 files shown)


JamesCropcho:
If you don't mind me asking, @omachala , I'd like to repoen this. It could be a good feature (what does everyone else think?

Also ( @todvora ) every test failure somehow amounts to

```
  13) Unnamed object analysis should use different array escape key:
     SyntaxError: Unexpected token u in JSON at position 113
      at JSON.parse (<anonymous>)
      at Tester._callee4$ (spec/utils/Tester.js:47:35)
      at tryCatch (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
      at step (spec/utils/Tester.js:23:191)
      at spec/utils/Tester.js:23:361
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:160:7)
```

…which I do not understand, although I've only given it a cursory glance.

@omachala , perhaps you could remedy the test failures and we can accept this pull request?


James

omachala:
Hi @JamesCropcho as a reopen reaction I tried fix the tests and extend BasicAnalyticsTest by including 'lastValue' option. Please keep in mind the feature is experimental.

todvora:
Hi all,
The feature looks good to me. Big plus for extending the test suite as well! Thank you for your contribution Ondrej.

All tests are passing on Travis-CI. Where do you see the exception James? Is it still occurring?

Maybe some edge cases to discuss:

- There is a dot used as undefined / unknown value. Will the meaning be clear to users?
- The logic supports only Boolean, String, Number, ObjectId and Date. What about [other types](https://docs.mongodb.com/manual/reference/bson-types/)? Like binary data?

Thanks,
Tomas

omachala:
Tomas (@todvora), 

thanks for your feedback. Dots was good solution for my usage, but you are absolutely right that the meaning for other users is unclear, so I changed unresolved values into combination of square brackets + type. In the case of other data types support I made some small improvements based on varietyTypeOf() possible outputs and print binary data in its hex form. For example output see doc.

JamesCropcho:
Ondrej ( @omachala ) this is a pretty fantastic pull request. Useful feature, enhanced test coverage, documentation…nice.

@todvora, the exception I mentioned was from the test suite. So naturally, yes, it is resolved.

I am happy to accept this pull request. On behalf of the many users of Variety around the world, thank you, Ondrej.


James

PR #148, opened 2019-01-28
Show type of array values

michaelmior:
This is a bit of a placeholder, but the gist is that it would be helpful to show the types of array elements.
I suspect this might be better handled by showing the top-level type as simply `Array` as is done currently, and then show something like `foo.XX` with the type of each element.

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -152,7 +152,11 @@ Released by Maypop Inc, © 2012-2018, under the MIT License. */
       return typeofThing[0].toUpperCase() + typeofThing.slice(1);
     } else {
       if (thing && thing.constructor === Array) {
-        return 'Array';
+        if (thing.length > 0) {
+          return 'Array-' + varietyTypeOf(thing[0]);
+        } else {
+          return 'EmptyArray';
+        }
       } else if (thing === null) {
         return 'null';
       } else if (thing instanceof Date) {


todvora:
Hi @michaelmior,
Thank you for your MR!

The problem I see with this approach is that an array may contain different types of values. If we judge the whole array based only on the first element (`varietyTypeOf(thing[0])`), we may provide a misleading information. 

There is already a bit similar functionality implemented: https://github.com/variety/variety#include-last-value. Could we reuse / build on that?

Thanks,
Tomas

michaelmior:
Absolutely that's a problem which I tried to identify in my description above. I think what could perhaps be reused more useful is type determination for nested objects since there's already a mechanism to show the multiple different types which exist. One approach could just be to find a way to treat the elements of an array like any other property of an object and use the same logic there.

michaelmior:
@JamesCropcho Should I interpret this to mean that there's no interest in adding this functionality?

JamesCropcho:
Ah, pardon, Michael: maybe for once I should have actually been more verbose than less so.

I mostly closed this due to inactivity, but also to a degree because Tomáš (@todvora) did not seem keen on the implementation (independent of the interface).

Yes, as he said these "arrays" in question are pseudoarrays at best because they accept multiple types. _(Aside: they behave more like either doubly-linked lists, or like a hash-map which only accepts integer keys—I haven't looked at the implementation so I don't know, but at least, they are not arrays)._

The implementation in this PR is written for arrays, and so it will often report an inaccurate result for MongoDB's "psuedoarray type."

I do not see a way around that. Do you, Michael or Tomáš? Does anyone else?

michaelmior:
Thanks for clarifying. I think the specific implementation probably does need work. But I think the existing type inference stuff can be used to guess the types used. Anyway, since we're all agreed that the current implementation is not that useful, I'll leave this closed unless I end up coming up with something better.

todvora:
Hi all,
Thank you for sharing your view. I agree - the implementation could provide misleading information and possibly lead to wrong decisions. So I would be worried to merge this feature as it is.

But @michaelmior please feel free to reopen it or discuss any other ideas you have to this topic.

Thank you for your contribution and your time!

Best regards,
Tomas

Issue #168, opened 2023-05-25
Errors running Variety with mongosh + DocumentDB


CodyDWJones:
I had to make some changes for Variety to run with the new MongoDB shell.

* `__quiet` is no longer defined. While I couldn't find a way to determine whether `--quiet` was passed on the command line, that option's description is "Silence output from the shell _during the connection process_" which suggests it should not be overloaded to control the script's logging too.
* `tojson()` is no longer defined.
* `db.getSisterDB()` no longer exists. `getSiblingDB()` should be used instead.
* Running the script logs the warning:
```
DeprecationWarning: Collection.count() is deprecated. Use countDocuments or estimatedDocumentCount.
```
* Assigning `dbQuery.prototype.reduce` fails. This function is only used once, so the code would be just as simple if the function body were inlined.

PR #169, opened 2023-05-26
PR for #168 - make Variety work with mongosh

CodyDWJones:
It seems a JS script can no longer check if `mongosh` was started with the `--quiet` flag, so this commit changes Variety to be always verbose.

A `quiet` config option could be added that would be distinct from the CLI's own `--quiet`, though the circular dependency between `log()` and `readConfig.read()` would need to be broken.


javespi:
Ty @CodyDWJones for the PR

Thanks to your fork I was able to get schemas in a Mongo 7 database

PR #218, opened 2026-04-03
Expand typed linting and checkJs coverage

JamesCropcho:
## Summary

Add type-aware ESLint coverage for `spec/utils` and widen the existing `checkJs` pass from the helper layer to the full `spec/` tree.

## What changed

- add local `typescript-eslint` and `@types/mocha` dev dependencies
- add a type-aware ESLint block for `spec/utils/**/*.js` backed by `tsconfig.checkjs.json`
- widen `tsconfig.checkjs.json` from `spec/utils` to all of `spec` and add Mocha types
- tighten helper and test JSDoc/types so the wider `checkJs` run passes cleanly
- annotate the sample CSV plugin so it can stay inside the widened typecheck surface
- update the README to describe the broader `typecheck` scope and the new typed linting on `spec/utils`

## Why

This combines the next two modernization steps in the planned sequence: introducing typed linting where the helper layer already had strong JSDoc, then widening `checkJs` to the rest of the spec code so more of the Node-side JavaScript gets static coverage.

## Impact

The repo now gets type-aware ESLint feedback for `spec/utils` and `tsc --checkJs` coverage across the full `spec/` tree. That raises the floor for test/helper code without crossing into the shell-executed runtime files yet.

## Validation

- `npm run lint`
- `npm run typecheck`
- `./node_modules/.bin/markdownlint README.md`
- `npm run test:mocha`

## Test notes

`npm run test:mocha` is blocked in this sandbox because localhost MongoDB connections fail with `connect EPERM` to `127.0.0.1:27017` / `::1:27017`, so I treated that as an environment limitation rather than a change-specific regression signal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,9 +344,9 @@ Pre-commit hooks are managed by [Husky](https://typicode.github.io/husky/) and i
 - `npm run lint:yaml` — js-yaml (YAML files)
 - `npm run lint:dockerfile` — hadolint (`docker/Dockerfile.template`)
 - `npm run lint:shell` — shellcheck (shell scripts)
-- `npm run typecheck` — TypeScript `checkJs`/JSDoc validation for Node-side helper code under `spec/utils`
+- `npm run typecheck` — TypeScript `checkJs`/JSDoc validation for Node-side spec code under `spec`
 
-ESLint applies a shared baseline of formatting and safety rules across the repo. That shared baseline now also bans repo-specific legacy patterns such as `Function('return this')`, `indexOf(...)` presence checks, and unguarded `for...in` loops. Node-side JavaScript such as `eslint.config.js`, the test suite, and `spec/utils` also opts into a stricter modernization set (`const`, template literals, object shorthand, `Object.hasOwn`, and throwing `Error` objects). Shell-executed files such as `variety.js` and shell plugins intentionally stay on the shared baseline until the project explicitly drops legacy `mongo` shell compatibility.
+ESLint applies a shared baseline of formatting and safety rules across the repo. That shared baseline now also bans repo-specific legacy patterns such as `Function('return this')`, `indexOf(...)` presence checks, and unguarded `for...in` loops. Node-side JavaScript such as `eslint.config.js`, the test suite, and `spec/utils` also opts into a stricter modernization set (`const`, template literals, object shorthand, `Object.hasOwn`, and throwing `Error` objects). The `spec/utils` helper layer now also uses type-aware `typescript-eslint` rules backed by `tsconfig.checkjs.json`. Shell-executed files such as `variety.js` and shell plugins intentionally stay on the shared baseline until the project explicitly drops legacy `mongo` shell compatibility.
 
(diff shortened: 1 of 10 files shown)


JamesCropcho:
This is a tooling/quality PR — extending TypeScript `checkJs` and type-aware lint rules from `spec/utils` to the full `spec` directory. Here's my read:

**What it does:**
- Adds `typescript-eslint` with `recommendedTypeChecked` rules scoped to `spec/utils/**/*.js`
- Expands `tsconfig.checkjs.json` from `spec/utils` to all of `spec`, adds `@types/mocha` so `describe`/`it` etc. are typed
- Adds JSDoc type annotations and null guards across spec files to satisfy the new strictness
- Adds `lastValue` parameter to `JsonValidator.validate()`

**Solid choices:**
- `filter((match) => match !== null)` over `filter(match => match)` is a real improvement — clearer intent and satisfies type-aware rules
- `typeof error.code === 'number' && error.code > 0` is stronger than the original bare `err.code > 0` — narrows the type before using it
- Adding `@types/mocha` is the right fix when expanding checkJs coverage to test files that use `describe`/`it`

**Two things worth flagging:**

1. **`tseslint.configs.recommendedTypeChecked[2].rules`** — accessing the preset by index `[2]` is fragile. If `typescript-eslint` reorganizes its config array between versions, this silently picks up different (or no) rules. Spreading the full `recommendedTypeChecked` array as a config entry, or using a named export if available, would be more robust.

2. **`arguments.length === 5` in `JsonValidator.validate()`** — works, but it's an archaic pattern for optional parameter detection. A cleaner signal would be a distinct sentinel (e.g. `const UNSET = Symbol()`) or splitting into a separate `validateWithLastValue()` method. Not a blocker, but worth a second look given this is utility code that others will build on.

3. **JSDoc indentation in `validate()`** — the `@param` tags in the updated JSDoc are indented inconsistently with the rest of the class. Cosmetic, but easy to fix.

**Bottom line:** The coverage expansion is worthwhile and the individual type fixes are genuine improvements. The index-based preset access is the one thing I'd want addressed before merging — it's a latent fragility in the lint setup.

JamesCropcho:
Thanks, this is helpful.

I agree that `tseslint.configs.recommendedTypeChecked[2].rules` is the one substantive concern here. Pulling rules by array index is brittle, and I wouldn’t want to merge the typed-lint setup with that latent dependency on preset ordering. I’m planning to switch that to a config composition that doesn’t rely on `[2]` while still keeping the type-aware rules scoped to `spec/utils`.

On `JsonValidator.validate()`, `arguments.length === 5` was meant as a compatibility-preserving bridge so the existing call sites could stay on one helper while adding `lastValue` assertions. I agree it’s not an especially elegant long-term shape, but I treated it as non-blocking compared with the lint-config fragility. If we keep touching that utility after this PR, I’d rather either split the API or introduce an explicit sentinel than keep leaning on `arguments`.

I also agree on the JSDoc indentation point. That’s an easy cleanup, and I’ll fold it into the follow-up adjustment so the utility reads cleanly as well as typechecks cleanly.

PR #230, opened 2026-04-13
Drop legacy mongo shell; require mongosh

JamesCropcho:
## Why

The legacy `mongo` shell has been end-of-life for years:

| Event | Date |
|---|---|
| `mongo` **deprecated** | MongoDB 5.0 (July 2021) |
| `mongo` **removed** from server distribution | MongoDB 6.0 (July 2022) |
| MongoDB 4.x (last version to ship `mongo`) **EOL** | February 2024 |
| MongoDB current stable | **8.x** (as of 2026) |

MongoDB 6.0+ does not ship `mongo` at all. The only users who would have it are on MongoDB 4.x, which has been unsupported for over two years. The practical overlap of "using variety.js" + "on EOL MongoDB" + "hasn't installed `mongosh` separately" is vanishingly small.

Additionally, `mongo` and `mongosh` are not fully equivalent: they differ in BSON type handling (e.g. `mongosh` promotes BSON `Double` and `Int32` to plain JS numbers — already noted in `DatatypeRecognitionTest.js`). Keeping the fallback means silently running a different code path with different semantics and zero test coverage.

## What changed

- **`bin/variety`** — removed the `elif command -v mongo` branch; the script now errors immediately with a clear message if `mongosh` is not in `PATH`
- **`spec/utils/MongoShell.js`** — removed the runtime shell-detection function (`findMongoShellCommand`) and its memoising promise; replaced with a direct `execFile('mongosh', …)` call
- **`eslint.config.js`** — removed the now-obsolete comment that deferred applying `nodeModernizationRules` to shell-executed files pending this drop
- **`README.md`** — updated all five spots that described `mongo` as a fallback or still-viable option; added a note that `variety.js` and plugins can now be opted into the modernization ruleset in a follow-on PR

## What's not in this PR

- Opting `variety.js` and shell plugins into `nodeModernizationRules` — that's a straightforward follow-on now that the blocker is removed
- A CI matrix leg that ever exercised the `mongo` path — there was none; the fallback was dead code in testing

## Test plan

- [ ] Pre-commit hooks (ESLint, shellcheck, tsc) pass — verified locally
- [ ] `npm run test:docker` against MongoDB 7.0 and 8.0 (both ship `mongosh` only)
- [ ] Confirm `bin/variety` exits 127 with a clear error when `mongosh` is absent

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,7 @@ So, let's see what we've got here:
 
 _(`test` is the database containing the collection we are analyzing.)_
 
-These examples use `mongosh`. If your environment still ships the legacy `mongo`
-shell instead, substitute that executable in the commands below.
+These examples use `mongosh`, which is required to run Variety.
 
 Hmm. Looks like everybody has a `name` and `_id`. Most, but not all have a `bio`.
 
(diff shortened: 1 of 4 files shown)


JamesCropcho:
Closing this draft PR in favor of issue #293, which tracks this work going forward.

PR #240, opened 2026-04-14
Delineate implementation and interface sections in variety.js

JamesCropcho:
## Summary

Prepares for an eventual split of `variety.js` into multiple files by organizing it into two clearly-bounded IIFEs with a documented boundary between them.

- **IMPLEMENTATION SECTION** — pure, transport-agnostic analysis logic. Functions take their dependencies (`config`, and where needed a `log` function or a `deps` bag) as explicit parameters. Exports a bundle via `shellContext.__varietyImpl`, which the interface section consumes and then deletes.
- **INTERFACE SECTION** — everything that touches shell globals: input reading (`collection`, `plugins`, `__quiet`, `slaveOk`, etc.), input validation, `readConfig`, `createPluginsRunner`, and the call to `impl.run()`.

Notable mechanical changes:
- `mergeDocument` / `reduceDocuments` now take `log` as an explicit 4th argument instead of closing over the shell-global `log`.
- `createAsciiTable` now uses its `results` parameter for `maxDigits` rather than the outer `varietyResults` binding (identical value in every call path; required for the future split).
- `filter` / `comparator` renamed `buildResultFilter` / `compareResults` and moved into the implementation section.
- `run(config, pluginsRunner, deps)` is a new single entry point in the implementation section; the interface section constructs and passes the deps bag.

## Known limitations / honest caveats

**`shellToJson` placement.** Under the strict rule "implementation = pure helpers only, interface = touches shell globals", `shellToJson` belongs in the interface section — it reads the shell globals `tojson` and `EJSON`. It lives in implementation to avoid duplication or threading it through `getBinDataHex`, but this is the weakest seam in the split and worth revisiting when the actual file split happens.

**`__varietyImpl` handoff.** The two IIFEs share state via a `shellContext.__varietyImpl` property that is `delete`d at the end, keeping the pair idempotent and leaving no global pollution after execution. It works, but it is a shell-specific kludge that previews how the eventual two-file version will communicate — not pretty, but honest about the constraint.

**"Pure" is aspirational.** `getSpecialTypeName` references `NumberLong`, `ObjectId`, and `BinData` via `typeof` guards; `shellToJson` reads `tojson` / `EJSON`; `getBinDataHex` reads `Buffer`. The implementation section degrades gracefully when these are absent rather than being genuinely transport-agnostic. Proving true portability would require exercising the implementation in plain Node with a mock `db` — that's left for a future step.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run typecheck` — clean
- [x] All 35 mocha integration tests pass unchanged against MongoDB 8.0 (no test modifications required)

### DIFF
--- a/variety.js
+++ b/variety.js
@@ -17,27 +17,36 @@ Released by James Cropcho, © 2012–2026, under the MIT License. */
 // for…of, Object.keys()/Object.entries(), and class. Later additions such
 // as Object.hasOwn() (ES2022) are absent and must not be used here.
 // See eslint.config.js for the enforced rule set.
+
+// -----------------------------------------------------------------------------
+// This file is organized in two sections, in preparation for an eventual split
+// into separate files:
+//
+//   1. IMPLEMENTATION SECTION — pure, transport-agnostic analysis logic.
+//      Functions take their dependencies (config, and where needed a `log`
+//      function or a `deps` bag holding shell primitives) as explicit
+//      parameters. The section hands a bundle of functions to the interface
+//      section via `shellContext.__varietyImpl`.
+//
+//   2. INTERFACE SECTION — everything that touches shell globals: reading
+//      input (`collection`, `plugins`, `__quiet`, `slaveOk`, etc.), the
+//      config-echo logging, plugin loading via `load()`, input validation,
+//      and constructing the dependency bag passed to `impl.run()`.
+//
+// The handoff property is deleted at the end so the pair is idempotent and
+// does not pollute the shell's global namespace after execution.
+// -----------------------------------------------------------------------------
+
+
+// =============================================================================
+// IMPLEMENTATION SECTION
+// =============================================================================
 (function (shellContext) {
-  'use strict'; // wraps everything for which we can use strict mode ―JC
+  'use strict';
 
   shellContext = typeof globalThis !== 'undefined' ? globalThis : shellContext;
 
-  const shellIsQuiet = () => {
-    if (typeof __quiet !== 'undefined' && __quiet) {
-      return true;
-    }
-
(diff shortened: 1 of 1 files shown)


JamesCropcho:
I think this is a good direction overall.

`variety.js` is carrying a lot of different responsibilities now, so making the analysis/core-vs-shell/interface boundary explicit before doing a true file split feels worthwhile. The part I like most in this attempt is that it is not just cosmetic reordering: passing `log` explicitly, introducing `run(config, pluginsRunner, deps)`, and moving the filter/sort helpers into the implementation section all help make the seam real.

On this particular PR, the refactor also reads as careful rather than overly clever. I walked the patch with an eye toward regressions and didn’t see behavioral changes I’d call scary, and I verified `npm run lint` and `npm run typecheck` locally on the PR branch. I also tried `npm test`; that got through lint and the container build, then failed in the harness with `/opt/variety/docker/init.sh: Permission denied`. Since this PR only touches `variety.js`, I don’t read that as change-specific signal either way.

The main bigger-picture caveat for me is that the temporary architecture only really pays off if it leads to a next step fairly soon. The `__varietyImpl` handoff is a reasonable transitional hack, but it is still a hack, and the “implementation section is pure” story is intentionally not quite true yet (`shellToJson`, BSON wrappers, `Buffer`, etc.). So my take is:

- good idea in principle
- solid execution for an intermediate refactor
- worth merging if it is genuinely setting up the next split / testability step
- somewhat more ceremony than I’d want as a long-term resting state

So: no blockers from me, and I think this is a net improvement, especially as a stepping stone. If there is a follow-up issue/PR to actually split files or add direct tests around the implementation layer, that would make the value proposition here much stronger.

PR #242, opened 2026-04-14
Split variety.js into src/ sources with a build step

JamesCropcho:
## Summary

- Move the implementation and interface sections of `variety.js` into `src/impl.js` and `src/interface.js`.
- Add `build.js` to concatenate the two sources underneath a generated-file banner (`npm run build`).
- Commit the generated `variety.js` so `mongosh variety.js` works from a fresh clone without a build step; CI runs `npm run verify:build` to fail if the committed file drifts from its sources.
- Extend the existing shell-compatible ESLint ruleset (previously scoped to `variety.js`) to cover the `src/*.js` sources too.
- Document the split and build workflow in `README.md`.

## Test plan

- [x] `npm run build` regenerates `variety.js` byte-for-byte
- [x] `npm run verify:build` passes on a clean tree and fails (with a useful message) when a source file drifts from `variety.js`
- [x] `npm run lint`, `npm run lint:json`, `npm run lint:markdown`, `npm run lint:yaml`, `npm run typecheck` all pass
- [x] `node --check` parses each of `variety.js`, `src/impl.js`, `src/interface.js`
- [ ] CI `test` / `test-smoke` jobs exercise the built `variety.js` against MongoDB 5.0 / 7.0 / 8.0


JamesCropcho:
Overall I think this is a well-executed refactor. The split between transport-agnostic logic (`src/impl.js`) and shell-facing glue (`src/interface.js`) is clear, keeping the generated `variety.js` committed preserves the fresh-clone `mongosh variety.js` workflow, and adding `verify:build` in CI is the right backstop for drift.

A couple of small omissions / nits:

- Coverage is not completely uniform across the new JS files. `src/impl.js` and `src/interface.js` are in good shape: `npm run lint` already hits them via `eslint .`, and this PR explicitly adds them to the shell-compatible rule set in `.eslint.config.js`. `build.js` is also linted by the generic `files: ['**/*.js']` block, but it does **not** get the stricter Node-side modernization rules (those still only target `.eslint.config.js` and `spec/**/*.js`), and it is also outside `.tsconfig.checkjs.json`'s `include` set. So the answer to “do all the preexisting JS checks cover the new files, including `./build.js`?” is: partially, but not all of them.
- `verify:build` is enforced in GitHub Actions now, but not in `.husky/pre-commit`, so a stale committed `variety.js` can still make it into a local commit and only fail later in CI. Not a blocker, but it feels like the one missing guardrail in an otherwise thoughtful workflow.
- The new README section is good. I’d probably also add `npm run build` / `npm run verify:build` to the nearby quick-reference/tooling lists so contributors see those commands in the same scan path as the other repo checks.

So: generally strong, with a couple of small workflow-consistency gaps rather than anything fundamentally wrong with the refactor itself.

JamesCropcho:
All three addressed in the amended commit:

- **`build.js` coverage** — added to both the `nodeModernizationRules` block in `.eslint.config.js` and the `include` list in `.tsconfig.checkjs.json`, so it gets the same stricter Node-side lint and type-checking as `.eslint.config.js` and the spec tree.
- **Pre-commit guardrail** — `npm run verify:build` is now the first step in `.husky/pre-commit`, so a stale `variety.js` is caught locally before it reaches CI.
- **README tooling list** — `npm run verify:build` added to the pre-commit hook bullet list; the `typecheck` entry updated to mention `build.js` explicitly.

PR #247, opened 2026-04-15
Evolve bin/variety towards being a real CLI

### LINKED ISSUE #232: Consider making `variety` the primary documented interface
## Summary

Variety currently documents the raw Mongo shell form as the default way to run it:

```bash
mongosh test --eval "var collection = 'users'" variety.js
```

That is explicit and powerful, but it is also a fairly low-level UX. Since this repo already ships a `bin/variety` wrapper and publishes it as the package's `variety` executable, I think it is worth considering whether `variety` should become the primary documented interface going forward, with the raw `mongosh ... variety.js` form retained as the lower-level / advanced form.

I am intentionally framing this as a design and maintenance question, not as "we must definitely change the docs immediately."

## Why this feels timely

PR #227 removed the old README references that pointed users to the archived `variety-cli` companion project:

- #227
- https://github.com/variety/variety-cli

That cleanup was correct. `variety-cli` is archived and should not be re-promoted as the recommended way to use Variety.

At the same time, removing those references also leaves us with a clearer question:

- should Variety continue to primarily document the raw `mongosh ... variety.js` invocation?
- or should the first-party `variety` executable become a more fully featured CLI and then become the default documented entrypoint?

## Current state

Today we effectively have three "interaction models":

1. Raw shell invocation, which is what most of the README uses:
   ```bash
   mongosh test --eval "var collection = 'users'" variety.js
   ```

2. The current first-party wrapper, which is documented in the README after #227:
   ```bash
   DB=test EVAL_CMDS="var collection = 'users', outputFormat='json'" VARIETYJS_DIR=. bin/variety
   ```

3. The old archived `variety-cli` syntax, which is no longer recommended, but had a much nicer user-facing shape:
   ```bash
   variety test/users --outputFormat='json' --quiet
   ```

The raw shell form is powerful, but verbose.
The current wrapper exists, but still feels like a convenience shim rather than a first-class CLI.
The archived CLI had a much nicer surface area, but lives in an unmaintained repo and should not be revived as an external dependency.

## Why consider changing the default docs at all?

The current default examples ask users to understand several things at once:

- which Mongo shell binary to run (`mongosh` vs `mongo`)
- how to target a DB
- how to pass Variety settings through `--eval`
- how to point at `variety.js`
- how to reason about shell quoting

That is manageable for experienced users, but not especially friendly.

By contrast, a primary interface centered on `variety` could be much more approachable, while still compiling down to the same underlying execution model.

This repo already owns the `variety` executable name, so there is no naming problem to solve. The main question is whether we want to invest in that UX and then promote it.

## Why the current `bin/variety` wrapper is probably not ready to be the default yet

The current wrapper is useful, but it still exposes a fairly low-level interface:

- `DB`
- `EVAL_CMDS`
- `VARIETYJS_DIR`

That means users still have to write JavaScript assignments inside `EVAL_CMDS`, and the docs still end up explaining the mechanics instead of just showing a straightforward command-line experience.

So I do not think the right move is simply "replace README examples with the current wrapper as-is."

Instead, I think the real question is whether `bin/variety` should evolve into a proper first-party CLI.

## The archived `variety-cli` project is still useful as design inspiration

Again, I do **not** think we should reintroduce `variety-cli` as a dependency or tell users to install it.

However, its syntax seems like a very good starting point for a first-party CLI design.

From the archived repo:

- Repository: https://github.com/variety/variety-cli
- Basic syntax:
  ```bash
  variety db_name/collection_name [options]
  ```

Examples from its README and manpage:

```bash
variety test/users --outputFormat='json' --quiet
variety logs/webserver --sort='{"created":-1}' --limit=100
variety test/users --query='{"bio":{"$exists":true}}'
variety test/users --host localhost --port 27017
variety test/users --plugins='/home/user/variety-plugin-csv.js'
```

A few especially good ideas from that old interface:

- `db/collection` as the main positional target
- direct flags for Variety options instead of forcing everything through `--eval`
- pass-through support for Mongo shell flags like `--host`, `--port`, `--username`, `--password`, `--quiet`
- a much smaller quoting and bootstrapping burden in normal usage

The old CLI also used HJSON for `--query` and `--sort`, which is interesting as inspiration, though that part is more debatable.

## Possible direction

Consider turning the current first-party `variety` executable into a real CLI with a surface something like:

```bash
variety test/users
variety test/users --outputFormat=json --quiet
variety test/users --limit=100 --maxDepth=3
variety test/users --query='{"bio":{"$exists":true}}'
variety test/users --host localhost --port 27017
```

Under the hood, it could still:

- prefer `mongosh` when available
- fall back to legacy `mongo` where needed
- translate CLI flags into the same Variety globals and shell invocation model already used today

That would let us preserve the implementation model while improving the UX.

## Questions worth deciding

- Do we want `variety` to become a first-class CLI, or do we prefer to keep the raw shell invocation as the canonical interface?
- If we do improve the CLI, should `db/collection` be the positional syntax?
- Should common Mongo shell flags be passed through directly?
  - `--host`
  - `--port`
  - `--username`
  - `--password`
  - `--authenticationDatabase`
  - `--quiet`
- Should `--query` and `--sort` accept strict JSON only, or something more forgiving?
- Should the executable resolve `variety.js` relative to the installed package automatically, so users do not have to think about file paths?
- Once the CLI is good enough, should the README use `variety ...` for most examples and move raw `mongosh ... variety.js` usage to an advanced section?

## My bias

My bias is:

- **do not** switch the docs default to the current env-var wrapper unchanged
- **do** consider evolving the first-party `variety` executable into a real CLI
- if that happens, the old archived `variety-cli` syntax is a very good starting point for the UX, even though the archived project itself should stay archived and un-recommended

## Possible acceptance criteria

If we decide to pursue this, a good end state might look like:

- `variety test/users` is a supported, documented command
- common Variety options are exposed directly as CLI flags
- common Mongo shell connection and auth flags are passed through
- `--quiet` works cleanly with JSON output
- the executable resolves the bundled `variety.js` reliably
- README examples mostly use `variety ...`
- raw `mongosh ... variety.js` examples remain documented for advanced, transparent, or debugging use cases
- the CLI behavior is covered by tests

## Non-goals

- Recommending the archived `variety-cli` project again
- Copying the old implementation wholesale
- Removing support for direct `mongosh ... variety.js` usage

JamesCropcho:
Closes #232.

## Summary
- evolve the bash-only `bin/variety` wrapper into a real Node CLI that supports `variety DB/COLLECTION [options]`
- preserve the documented `DB` / `EVAL_CMDS` / `VARIETYJS_DIR` compatibility mode while keeping raw `mongosh ... variety.js` usage intact
- add focused CLI parsing and wrapper coverage, and update README/CONTRIBUTING plus lint/package metadata for the new entrypoint

## Testing
- `npm run lint`
- `npm run lint:markdown`
- `npm run typecheck`
- `npm run verify:build`
- `./node_modules/.bin/mocha --reporter spec --timeout 15000 spec/BinWrapperTest.js spec/CliOptionsTest.js`
- `npm run test:docker`
- `MONGODB_VERSION=5.0 npm run test:docker`

### DIFF
--- a/.eslint.config.js
+++ b/.eslint.config.js
@@ -55,7 +55,7 @@ module.exports = [
   { ignores: ['.claude/'] },
   js.configs.recommended,
   {
-    files: ['**/*.js'],
+    files: ['**/*.js', 'bin/variety'],
     languageOptions: {
       sourceType: 'script',
       globals: {
@@ -79,7 +79,7 @@ module.exports = [
     },
   },
   {
-    files: ['.eslint.config.js', 'build.js', 'spec/**/*.js'],
+    files: ['.eslint.config.js', 'bin/variety', 'build.js', 'lib/**/*.js', 'spec/**/*.js'],
     ignores: ['spec/assets/**/*.js'],
     rules: nodeModernizationRules,
   },
--- a/.tsconfig.checkjs.json
+++ b/.tsconfig.checkjs.json
@@ -21,6 +21,8 @@
     ]
   },
   "include": [
+    "bin/variety",
+    "lib/**/*.js",
     ".eslint.config.js",
     "build.js",
     "spec/**/*.js",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,7 +80,7 @@ Pre-commit hooks are managed by [Husky](https://typicode.github.io/husky/) and i
 - `npm run lint:yaml` — js-yaml (YAML files)
 - `npm run lint:dockerfile` — hadolint (`docker/Dockerfile.template`)
 - `npm run lint:shell` — shellcheck (shell scripts)
-- `npm run typecheck` — TypeScript `checkJs`/JSDoc validation for `.eslint.config.js`, `build.js`, and Node-side spec code under `spec`
+- `npm run typecheck` — TypeScript `checkJs`/JSDoc validation for `bin/variety`, `lib/**/*.js`, `.eslint.config.js`, `build.js`, and Node-side spec code under `spec`
 
 ### ESLint Rulesets
 
@@ -100,7 +100,7 @@ Node-side JavaScript such as `bin/variety`, `lib/**/*.js`, `.eslint.config.js`, 
 
 #### Checked Files
 
-`npm run typecheck` runs TypeScript `checkJs` over `.eslint.config.js`, `build.js`, and the Node-side spec code via `.tsconfig.checkjs.json`. The `spec` tree also uses type-aware `typescript-eslint` rules, while shell-executed fixtures under `spec/assets` stay on the shared baseline.
+`npm run typecheck` runs TypeScript `checkJs` over the published Node CLI surface (`bin/variety` plus `lib/**/*.js`), `.eslint.config.js`, `build.js`, and the Node-side spec code via `.tsconfig.checkjs.json`. The `spec` tree also uses type-aware `typescript-eslint` rules, while shell-executed fixtures under `spec/assets` stay on the shared baseline.
 
 #### Extra Strictness
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
 - `src/interface.js` — the shell-facing layer that reads shell globals
   (`collection`, `plugins`, `slaveOk`, etc.), loads plugins, and hands
   dependencies to `impl.run()`.
-- `bin/variety` — the published Node entrypoint that implements the first-party
+- `bin/variety` — the published Node entrypoint that implements the main
   CLI surface.
 - `lib/cli*.js`, `lib/mongo-shell.js` — Node-side CLI parsing, compatibility
   handling, and Mongo shell invocation helpers.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,10 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
 - `src/interface.js` — the shell-facing layer that reads shell globals
   (`collection`, `plugins`, `slaveOk`, etc.), loads plugins, and hands
   dependencies to `impl.run()`.
+- `bin/variety` — the published Node entrypoint that implements the first-party
+  CLI surface.
+- `lib/cli*.js`, `lib/mongo-shell.js` — Node-side CLI parsing, compatibility
+  handling, and Mongo shell invocation helpers.
 
 `build.js` concatenates those two files under a generated-file banner. Edit the sources in `src/`, then run:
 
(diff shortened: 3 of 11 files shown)


JamesCropcho:
Overall: this lands cleanly against the acceptance criteria in #232. The split into `lib/cli-options.js` (parsing), `lib/mongo-shell.js` (execution), and `lib/cli.js` (entry/error translation) is easy to follow, and the `CliUsageError` / `CliRuntimeError` pairing with distinct exit codes (2 for usage, 127 for missing shell, passthrough for shell status) is exactly what I'd hope for in a first-class CLI. Test coverage is solid — the new `CliOptionsTest.js` plus the expanded `BinWrapperTest.js` exercise the target parsing, connection flag pass-through, help/version short-circuit, and the preserved env-var compatibility path. The compatibility vs. CLI `scriptPath` split (relative `./variety.js` for the old env-var mode, `path.resolve(__dirname, '..', 'variety.js')` for the new CLI) is the right call to avoid regressing existing users while making the new surface path-independent.

A few observations, none blocking:

1. **`@ts-nocheck` on all three new `lib/` files.** The repo already runs `tsc --checkJs` and the spec files went to the trouble of adding JSDoc typedefs for these modules — it'd be a shame for the implementation itself to opt out. Worth a follow-up to add proper JSDoc so the public shape is checked end-to-end.

2. **`parsePositiveInteger` accepts `0`.** `/^\d+$/` allows zero; the name suggests otherwise. Probably harmless for `--limit 0` / `--port 0` / `--maxDepth 0` (mongosh will reject bad ports), but either tighten the regex or rename to `parseNonNegativeInteger`.

3. **`readOptionValue` rejects any value starting with `--`.** That means `--eval '--foo'` fails unless the user uses `--eval=--foo`. Fine in practice for `query`/`sort`/`limit`/etc., just a sharp edge on the `--eval` escape hatch specifically. Not worth fixing unless it comes up.

4. **Shell arg ordering puts `--host`/`--port` before the database positional and `--quiet`/auth flags after.** mongosh accepts both orderings, so this is fine, just slightly unusual if someone eyeballs the printed command.

5. **Help text only advertises the camelCase flag names** (`--maxDepth`, `--outputFormat`, `--authenticationDatabase`) even though `FLAG_ALIASES` also accepts the kebab-case forms. Intentional and probably right (one documented spelling), but worth being explicit if someone asks.

None of the above would hold up merging. Nice work — this is a meaningful UX jump for the package.

JamesCropcho:
Thanks for the thoughtful review. I pushed `8371564` to address `#1` and `#2`.

- `#1`: removed the `@ts-nocheck` markers from `lib/cli-options.js`, `lib/cli.js`, and `lib/mongo-shell.js`, added JSDoc annotations, and expanded `.tsconfig.checkjs.json` / `npm run typecheck` coverage so it now includes the published CLI surface (`bin/variety` plus `lib/**/*.js`). I also switched the spec-side typing to `typeof import(...)` so the focused CLI tests consume the real module surface instead of a handwritten facsimile.
- `#2`: split the integer parsing so `--limit` and `--maxDepth` continue to allow `0` intentionally, while `--port` now requires a positive integer. That preserves the existing `limit=0` behavior while making the helper naming/behavior line up.
- `#3`: now tracked separately in its own issue, so I left that out of this PR.

I reran `npm run lint`, `npm run lint:markdown`, `npm run typecheck`, `npm run verify:build`, and `./node_modules/.bin/mocha --reporter spec --timeout 15000 spec/BinWrapperTest.js spec/CliOptionsTest.js` after the change.

PR #285, opened 2026-04-17
Restructure test directory layout and rename test helpers

JamesCropcho:
### Test directory restructure
- Moves all test subject subdirectories (`analysis`, `cli`, `formatters`, `mongo-shell`, `persistence`, `plugins`) into a new `test/cases/` layer, separating runnable test suites from test infrastructure
- Renames `test/utils/` → `test/helpers/` to better describe its role

### Helper renames
- `MongoShellRunner.js` (was `MongoShell.js`) — clarifies it is an execution helper, not a test of the `mongo-shell/` source module
- `VarietyHarness.js` (was `Tester.js`), class `Tester` → `VarietyHarness`
- `AnalysisResultsValidator.js` (was `JsonValidator.js`), class `JsonValidator` → `AnalysisResultsValidator`

### Fixture renames and reorganisation
- `seed-data.js` (was `SampleData.js`) — "seed" is the standard database testing term
- `ascii-output-fixture.js` (was `ExpectedAscii.js`) — literal, no snapshot-test update semantics
- `fixtures/plugins/csv-plugin.js` (was `fixtures/CsvPlugin.js`) — moved into a `plugins/` subdir to distinguish mongosh-loaded CommonJS plugin stubs from Node-imported test data; kebab-case matches real-world plugin naming conventions

### Infrastructure renames
- `test-in-container.sh` (was `test-on-docker.sh`) — the script supports Podman too
- npm script `test:container` (was `test:docker`) to match

### Reference updates
- All import paths, JSDoc typedefs, `package.json`, `.github/workflows/ci.yml`, and `CONTRIBUTING.md` updated throughout

## Test plan

- [ ] `npm run test:container` passes with Docker or Podman
- [ ] `npm run test:mocha` passes against a local MongoDB on port 27017
- [ ] `npm run lint` passes (ESLint + shellcheck + markdownlint)
- [ ] `npm run typecheck` passes


JamesCropcho:
High-level take: I think this PR is directionally sound. It makes `test/` read more like a small test subsystem: runnable suites, helpers, fixtures, and bin scripts are now distinct concepts instead of being mixed together. The churn is mostly mechanical, so the main risk is naming/API stability rather than behavior.

Decision-by-decision:

- `test/tests/` layer: sound intent, slightly awkward name. Separating runnable test suites from test infrastructure is useful; `test/tests` does stutter a bit. If this lands as-is it is still understandable, but `test/suites` or `test/cases` would read cleaner to me.
- `test/utils/` -> `test/helpers/`: sound. These are test-only collaborators, not generic utilities.
- `MongoShell.js` -> `MongoShellRunner.js`: sound. The new name says this helper executes shell work; it is not itself a test of the `mongo-shell/` module.
- `Tester` -> `VarietyHarness`: sound. The class owns DB setup, cleanup, and Variety invocation, so “harness” is much more accurate than “tester.”
- `JsonValidator` -> `AnalysisResultsValidator`: sound. It validates parsed analysis result rows, not JSON syntax.
- `SampleData.js` -> `seed-data.js`: sound. This is database seed input, and kebab-case is friendlier for fixture files.
- `ExpectedAscii.js` -> `AsciiSnapshot.js`: mostly sound, but weaker than the others. “Snapshot” can imply snapshot-test update semantics; if this is just a static expected-output fixture, something like `expected-ascii-output.js` or `ascii-output-fixture.js` might be more precise. Not a blocker.
- `fixtures/plugins/csv-plugin.js`: sound. Moving plugin fixtures under `fixtures/plugins/` usefully distinguishes mongosh-loaded plugin files from Node-imported data fixtures, and kebab-case matches real-world plugin naming.
- `test-on-docker.sh` / `test:docker` -> `test-in-container.sh` / `test:container`: the container terminology is sound because the harness supports Podman too. I would not drop the `test:docker` npm script, though. That command is likely user/contributor muscle memory, and “Docker-backed” remains common shorthand even when Podman is supported. Best version in my view: make `test:container` canonical, but keep `test:docker` as a compatibility alias.
- CONTRIBUTING/CI reference updates: sound. This is contributor workflow/tooling, so `CONTRIBUTING.md` is the right place for the docs update.

So my vote: good direction and conceptually mergeable, with one strong recommendation before merge: keep `npm run test:docker` as an alias for `npm run test:container`. The only other possible adjustments I’d consider are `test/tests` -> `test/suites` and maybe replacing `AsciiSnapshot` with a more literal expected-output name, but those are taste-level rather than blockers.

JamesCropcho:
Thanks for the thorough review — really useful signal on all three open points.

**`test:docker` alias** — agreed, will add `"test:docker": "npm run test:container"` before merge. Muscle memory is a real cost.

**`test/tests/` stutter** — fair point. Happy to rename to `test/suites/` if you'd prefer — it reads cleanly and avoids the repetition. Say the word and I'll do it in this PR.

**`AsciiSnapshot`** — the concern about snapshot-test update semantics is well-taken. `expected-ascii-output.js` is unambiguous, though a bit verbose. I'd lean toward `ascii-output.js` as a middle ground — short, literal, no snapshot connotation — but happy to go with whichever of the three (`AsciiSnapshot`, `expected-ascii-output`, `ascii-output`) you prefer.

I'll wait for your steer on the two naming questions before pushing anything, then land the `test:docker` alias at the same time.

JamesCropcho:
Done — both changes are pushed:

- `test/tests/` → `test/cases/`
- `AsciiSnapshot.js` → `ascii-output-fixture.js` (kebab-case, matching the other fixture files)

On the `test:docker` alias: intentionally not adding it, per the author's preference that the full rename takes priority over muscle-memory compatibility.

PR description updated to reflect the final state.

PR #295, opened 2026-04-20
Fix markdown-only skip breaking required status checks in CI

JamesCropcho:
## Summary

- Matrix test jobs (`test`, `test-smoke`) had a job-level `if: only-markdown != 'true'` that caused two problems when a markdown-only PR landed:
  - The job name showed the raw `${{ matrix.mongodb-version }}` template string instead of the expanded value (matrix variables are only substituted when a job actually runs)
  - The expanded check names branch protection requires (e.g. `CI / Test (Node 22, MongoDB 5.0)`) were never reported, leaving those required checks permanently pending ("Expected — Waiting for status to be reported")
- Fix: move the `only-markdown` guard to the step level so the job always starts, skips all real work, and exits success — satisfying required checks without running tests
- Documents the rationale in `CONTRIBUTING.md` and in a code comment adjacent to the affected steps

## Test plan

- [ ] Open a markdown-only PR and verify all three matrix check names appear as green (not pending/skipped)
- [ ] Open a non-markdown PR and verify tests still run normally

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,27 +75,43 @@ jobs:
   test:
     name: Test (Node 22, MongoDB ${{ matrix.mongodb-version }})
     needs: changes
-    if: needs.changes.outputs.only-markdown != 'true'
+    # No job-level if: here — see note below. Steps are guarded individually.
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         mongodb-version: ['5.0', '7.0', '8.0']
     steps:
+      # NOTE: The only-markdown skip is applied at the step level rather than
+      # the job level. A job-level `if: only-markdown != true` causes GitHub
+      # Actions to skip the entire job, which: (a) leaves the job name showing
+      # the raw `${{ matrix.mongodb-version }}` template string (matrix
+      # variables are only expanded when the job runs), and (b) never reports
+      # the expanded check names (e.g. "Test (Node 22, MongoDB 5.0)") that
+      # branch protection requires, leaving those checks permanently pending.
+      # Step-level guards let the job start, do nothing, and exit success —
+      # satisfying required checks while still skipping the expensive work.
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        if: needs.changes.outputs.only-markdown != 'true'
       - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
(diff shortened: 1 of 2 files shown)


JamesCropcho:
I took another look at this from the angle of "is this a good idea?", not just "are the checks green?"

My verdict: this is a reasonable short-term fix and I do not think it creates a hidden path where code changes can appear to pass while going untested. The important safety property is the `only-markdown` gate: it only returns true when every changed path is an existing `.md` file with `M` status. Additions, deletions, renames, workflow edits, package/tooling edits, and source changes all fall back to full CI. PR #295 itself did full CI because it changed `.github/workflows/ci.yml`, and PR #294 is a useful later example of the intended markdown-only behavior: lint still ran markdown lint, while the test/smoke steps were visibly skipped inside the job details.

I also checked the active ruleset for `main`; it requires the exact `Lint`, `Test (Node 22, MongoDB ...)`, and `Smoke Test` contexts. Given that setup, the old job-level `if:` is genuinely awkward because skipped matrix jobs do not emit the expanded required check names. So this PR is solving a real branch-protection/reporting problem, not papering over a failing test.

The fairness caveat: in the PR checks list, a markdown-only PR will show green `Test (...)` checks even though the test commands did not run. That is slightly ambiguous if someone reads the green check as "tests executed and passed." I would not call it misreporting, because the underlying reason is visible in the job steps and the detector is deliberately narrow, but it is not as semantically clean as a separate required aggregate check like `CI result` with the individual test jobs allowed to be skipped honestly.

So: safe enough for the current ruleset, and better than the previous pending-required-check behavior. If we want to make it feel more honest in the UI, I would add an explicit no-op step such as `Skip tests for markdown-only change` to the test and smoke jobs, or eventually move to a single required aggregate status and stop requiring each matrix job directly.

JamesCropcho:
I’m reverting this via #297 before replacing it with a different shape.

The reason is reporting semantics, not that #295 made code changes unsafe. The `only-markdown` detector is narrow enough that source/tooling/workflow changes still run full CI, but the workaround here makes required `Test (...)` and `Smoke Test (...)` checks turn green on markdown-only PRs even though the actual test commands are skipped. That fixes the branch-protection deadlock, but the PR checks list can be read as “tests ran and passed,” which is less honest than I want.

The replacement I’m going to build from post-revert `main` is an aggregate required gate: let the test matrix jobs be truly skipped for markdown-only changes, then require a separate always-running gate that explicitly passes because either full CI passed or because the change is markdown-only and tests were intentionally skipped. That keeps branch protection workable while making the UI’s story clearer.

Issue #309, opened 2026-04-21
slaveOk is deprecated — remove or replace in variety.js


JamesCropcho:
## Background

`variety.js` currently checks for a `slaveOk` global variable and calls `db.getMongo().setSlaveOk()` when it is set (line ~579):

```js
if (typeof slaveOk !== 'undefined') {
  if (slaveOk === true) {
    db.getMongo().setSlaveOk();
  }
}
```

`setSlaveOk()` was deprecated in MongoDB 4.x and removed in MongoDB 5.0+. It is not available in modern `mongosh`.

## Why this surfaced now

While auditing which `variety.js` options are not yet exposed through the `bin/variety` CLI (tracked in #232), `slaveOk` came up as a candidate for PR1 of the parity work. The decision was made **not** to expose it as a first-class CLI flag precisely because it is a deprecated/dead feature on any stack running MongoDB 5.0+ or modern `mongosh`. Promoting it in the help text would be misleading.

## Modern replacements

The correct way to allow secondary reads in contemporary MongoDB tooling is:

- `rs.secondaryOk()` — shell helper, sets read preference to secondary for the current shell session
- `db.getMongo().setReadPref('secondary')` — programmatic equivalent

## Suggested resolution

- Remove the `slaveOk` block from `variety.js`, or replace it with an equivalent that uses `setReadPref('secondary')` under a better-named variable (e.g. `secondaryOk` or `readFromSecondary`).
- If a replacement variable is introduced, expose it as a CLI flag in `bin/variety` as part of the ongoing parity work (#232).
- Add a note in `CONTRIBUTING.md` or a changelog entry if the old `slaveOk` variable name is removed (it is a breaking change for any scripts that set it).

## Related

- #232 — CLI parity audit and roadmap

JamesCropcho:
Quick correction from the audit before implementation: the issue's overall direction is right, but one detail is not quite accurate on the currently tested stack.

On MongoDB 8.2 with `mongosh` 2.8.2, `db.getMongo().setSlaveOk` is still present as a function; it is not literally absent there. However, calling it fails with:

```text
MongoshDeprecatedError: [COMMON-10003] Setting slaveOk is deprecated, use setReadPref instead.
```

I also verified that `db.getMongo().setReadPref('secondary')` works in that same environment and changes the read preference from `primary` to `secondary`.

[...]

PR #310, opened 2026-04-21
Expose all variety.js options via the first-party CLI

### LINKED ISSUE #232: Consider making `variety` the primary documented interface

JamesCropcho:
## Summary

Closes the 12-option parity gap identified in issue #232 (@JamesCropcho).

The `bin/variety` CLI previously exposed only 6 of Variety's 18 configurable options. This PR adds the remaining 13 options (including `--plugins`) so that everything available via raw `mongosh ... variety.js` is now reachable through `variety DB/COLLECTION [options]`.

### New flags

- `--lastValue` — capture one representative value per key
- `--showArrayElements` — include `tags.XX`-style keys in output
- `--compactArrayTypes` — render `Array(Type)` instead of plain `Array`
- `--arrayEscape <value>` — custom escape for array index keys (default `XX`)
- `--excludeSubkeys <json>` — JSON array of dot-paths to skip
- `--logKeysContinuously` — stream keys as they arrive
- `--slaveOk` — allow secondary reads on a replica set
- `--plugins <path>` — load a plugin file (comma-separated for multiple)
- `--persistResults` — write results to MongoDB
- `--resultsDatabase <value>` — target DB for persisted results
- `--resultsCollection <value>` — target collection for persisted results
- `--resultsUser <value>` — MongoDB username for results database
- `--resultsPass <value>` — MongoDB password for results database

Kebab-case aliases (e.g. `--show-array-elements`, `--exclude-subkeys`) are accepted for all multi-word flags. `--excludeSubkeys` expects a strict JSON array.

### Documentation

README updated to use `variety DB/COLLECTION` as the primary example throughout all relevant option sections. A complete flag reference table has been added to the CLI section.

### Tests

CLI test suite expanded from 20 to 29 tests covering all new flags, kebab aliases, and `--excludeSubkeys` validation.

## Test plan

- [x] All 29 CLI unit tests pass (`mocha test/cases/cli/BinWrapperTest.js test/cases/cli/CliOptionsTest.js`)
- [x] ESLint, TypeScript checkJs, markdownlint, SPDX, build-verify all clean
- [x] Docker-backed integration suite runs via `npm run test:container` in CI

Closes #232

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,11 +253,11 @@ Passing an unrecognised value throws an error listing the valid options.
 
 A plugin is a `.js` file that exports a plain object with lifecycle hooks. Load one with the `plugins` option (comma-separated for multiple):
 
-    $ mongosh test --quiet --eval "var collection='users', plugins='/path/to/my-plugin.js'" variety.js
+    $ variety test/users --plugins /path/to/my-plugin.js --quiet
 
 Pass per-plugin configuration by appending `|key=value` after the path:
 
-    $ mongosh test --quiet --eval "var collection='users', plugins='/path/to/my-plugin.js|delimiter=;'" variety.js
+    $ variety test/users --plugins '/path/to/my-plugin.js|delimiter=;' --quiet
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for the full hook reference and a guide to writing and testing plugins.
 
@@ -269,16 +269,16 @@ Variety can also read that option and mute unnecessary output. This is useful in
     $ mongosh test --quiet --eval "var collection = 'users', outputFormat='json', sort = { updated_at : -1 }" variety.js
 
 ### Log Keys and Types As They Arrive Option
-Sometimes you want to see the keys and types come in as it happens.  Maybe you have a large dataset and want accurate results, but you also are impatient and want to see something now.  Or maybe you have a large mangled dataset with crazy keys (that probably shouldn't be keys) and Variety is going out of memory.  This option will show you the keys and types as they come in and help you identify problems with your dataset without needing the Variety script to finish.  
+Sometimes you want to see the keys and types come in as it happens.  Maybe you have a large dataset and want accurate results, but you also are impatient and want to see something now.  Or maybe you have a large mangled dataset with crazy keys (that probably shouldn't be keys) and Variety is going out of memory.  This option will show you the keys and types as they come in and help you identify problems with your dataset without needing the Variety script to finish.
 
-    $ mongosh test --eval "var collection = 'users', sort = { updated_at : -1 }, logKeysContinuously = true" variety.js
+    $ variety test/users --logKeysContinuously
 
 ### Exclude Subkeys
-Sometimes you inherit a database full of junk.  Maybe the previous developer put data in the database keys, which causes Variety to go out of memory when run.  After you've run the `logKeysContinuously` to figure out which subkeys may be a problem, you can use this option to run Variety without those subkeys.  
+Sometimes you inherit a database full of junk.  Maybe the previous developer put data in the database keys, which causes Variety to go out of memory when run.  After you've run the `logKeysContinuously` to figure out which subkeys may be a problem, you can use this option to run Variety without those subkeys.
 
     db.users.insertOne({name:"Walter", someNestedObject:{a:{b:{c:{d:{e:1}}}}}, otherNestedObject:{a:{b:{c:{d:{e:1}}}}}});
 
-    $ mongosh test --eval "var collection = 'users', sort = { updated_at : -1 }, excludeSubkeys = [ 'someNestedObject.a.b' ]" variety.js
+    $ variety test/users --excludeSubkeys '["someNestedObject.a.b"]'
 
     +-----------------------------------------------------------------+
     | key                         | types    | occurrences | percents |
@@ -300,7 +300,7 @@ Sometimes you inherit a database full of junk.  Maybe the previous developer put
 
 By default, Variety suppresses keys that end with an array index (e.g. `tags.XX`), because the parent key already captures the `Array` type. If you want to see the types of the values _inside_ primitive arrays — useful for verifying element-type consistency — set `showArrayElements` to `true`:
 
-    $ mongosh test --eval "var collection = 'users', showArrayElements = true" variety.js
+    $ variety test/users --showArrayElements
 
 For example, given documents like:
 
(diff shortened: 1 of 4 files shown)


JamesCropcho:
Closing in favour of a 3-PR approach (Option C). The single-PR approach was not the agreed plan: it included --slaveOk (excluded pending issue #309), used a JSON-array syntax for --excludeSubkeys instead of the agreed repeatable-flag pattern, and folded PR2/PR3 scope (persistence, plugins) into a single commit. Will be replaced by three focused PRs.

Issue #264, opened 2026-04-16
Add an interactive Mongo shell API for Variety


JamesCropcho:
## Problem

Issue #131 clarified that Variety can be loaded from an already-running Mongo shell only by defining globals such as `collection` before calling `load("/path/to/variety.js")`. That works, but it is still an execute-on-load script interface rather than a real interactive shell API.

A shell-native API would make Variety easier to use from `mongosh` or the legacy `mongo` shell during exploratory database work.

## Possible direction

Explore adding an API that can be loaded without immediately running analysis, such as one of these shapes:

```js
load("/path/to/variety-shell.js");
db.users.variety({ limit: 100, outputFormat: "json" });
```

or:

```js
load("/path/to/variety-shell.js");
variety("users", { limit: 100, outputFormat: "json" });
```

The exact API can change, but the important behavior is that loading the helper should register callable functionality rather than immediately requiring a `collection` global.

## Compatibility notes

The existing script-style interfaces should keep working:

```bash
mongosh test --eval "var collection = 'users'" variety.js
```

```js
var collection = "users";
load("/path/to/variety.js");
```

## Acceptance criteria

- Loading the new shell API does not require a `collection` global and does not immediately run analysis.
- Users can analyze a collection from an interactive shell call and pass normal Variety options.
- Existing `mongosh ... --eval ... variety.js`, `mongo ... --eval ... variety.js`, and `var collection = ...; load(...)` flows continue to work.
- The behavior is covered by focused tests, ideally through the existing Docker-backed Mongo shell harness.
- README and CONTRIBUTING are updated where needed.

Thanks again to @abrin for raising the original shell-loading question.

JamesCropcho:
Status update: PR #332 has now merged into `main`.

That change does not implement the interactive Mongo shell API yet, but it does extract a reusable internal analysis engine so future shell-API work can build on a shared core instead of the current execute-on-load shell orchestration.

Follow-up engine issues are now tracked separately:

- #333 — make `core/engine.js` runtime-agnostic rather than ambient-runtime aware
- #334 — tighten the exported surface of `core/engine.js`

So the remaining work on this issue is still the actual shell-facing API shape and adapter behavior, but the first internal seam is now in place.

JamesCropcho:
Status update: PR #339 has now merged into `main`.

[...]

Issue #263, opened 2026-04-16
Make Variety usable directly from Node.js applications


JamesCropcho:
Follow-up to #125 and #262.

## Background

Issue #125 asked whether Variety could be used from a Node.js application, such
as an API endpoint that returns the available keys in a MongoDB collection. PR
#262 documents the current supported workaround: call the packaged `variety`
CLI, request JSON output, and parse stdout.

That is useful, but it still treats terminal output as the integration API.

## Problem

Node.js applications that want Variety's schema analysis currently need to
spawn a subprocess, depend on a MongoDB shell being installed, parse command
output, and carefully sanitize any user-influenced options before they reach the
CLI. That can be awkward for application code and risky for web endpoints.

## Goal

Provide a supported way to use Variety from Node.js application code without
requiring consumers to shell out to the CLI or treat stdout as the primary API.

## Possible Direction

- expose a Node.js API that returns structured analysis data directly
- accept a MongoDB Node driver `Db` or `Collection`, or a small adapter around it
- keep the existing CLI as a wrapper over the same underlying analysis behavior
- preserve `--outputFormat json` for command-line automation while making it
  unnecessary for in-process Node.js callers
- document plugin, persistence, query, sort, limit, and depth behavior for the
  programmatic API

## Open Questions

- Should this live in the current package, a future `@variety/core`, or a future
  `@variety/node` package?
- Which Node.js and MongoDB driver versions should be supported?
- How should existing Mongo shell-specific options map to a driver-backed API?
- Should the API support progress callbacks or streaming results for large
  collections?

## Acceptance Sketch

- README includes a Node.js example that does not use `child_process`
- tests cover calling the API against a MongoDB collection and receiving
  structured results
- the CLI continues to work and, where practical, shares behavior with the new
  API layer

JamesCropcho:
Status update: PR #332 has now merged into `main`.

That change does not implement the Node.js programmatic API yet, but it does extract a reusable internal analysis engine so future API work can build on a shared core instead of shell-specific orchestration.

Follow-up engine issues are now tracked separately:

- #333 — make `core/engine.js` runtime-agnostic rather than ambient-runtime aware
- #334 — tighten the exported surface of `core/engine.js`

So the remaining work on this issue is still the actual Node.js-facing API and adapter design, but the first internal seam is now in place.

JamesCropcho:
Status update: PR #339 has now merged into `main`.

[...]

PR #332, opened 2026-04-23
Extract reusable internal analysis engine

JamesCropcho:
## Summary
- split the pure document-reduction and result-finalization logic out of `core/analyzer.js` into new `core/engine.js`
- keep `core/analyzer.js` focused on cursor traversal, optional persistence, and formatter dispatch
- update the build docs, generated `variety.js`, and direct engine tests around the new internal seam

## Why
- move Variety toward the reusable internal engine needed for #263 and #264
- make future Node.js and interactive Mongo shell APIs share one analysis core instead of duplicating logic

## Impact
- creates an internal reducer/finalization API that future work can reuse without shell globals or formatter side effects

## Validation
- `npm run lint`
- `npm run typecheck`
- `npm run verify:build`
- `npm run lint:markdown`
- `npm run lint:spdx`
- `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/analysis/AnalyzerPropertyFuzzingTest.js test/cases/analysis/DeprecatedBsonTypesTest.js test/cases/mongo-shell/AdapterValidationTest.js`
- Mongo-backed Podman validation from this worktree:
  - targeted previously failing suite slice passed under a controlled `podman run ... --volume ...:Z --entrypoint bash ...` invocation
  - full `npm run test:mocha` inside the same container image was mostly green but not fully stable from this `/tmp` worktree mount because the default container entrypoint path handling differs there; the failures were late, transient Mongo availability errors rather than analysis-engine assertion regressions

Refs #263 and #264.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,21 +14,25 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
 
 ## Repo Layout and the `variety.js` Build
 
-`variety.js` at the repo root is a generated file, assembled from four sources in this order:
+`variety.js` at the repo root is a generated file, assembled from five sources in this order:
 
 - `core/formatters/ascii.js` — built-in ASCII table formatter. Self-contained
   IIFE; registers an `ascii` factory on `shellContext.__varietyFormatters`.
 - `core/formatters/json.js` — built-in JSON formatter. Self-contained IIFE;
   registers a `json` factory on `shellContext.__varietyFormatters`.
-- `core/analyzer.js` — pure, transport-agnostic analysis logic. No runtime
-  dependencies on shell globals, Node I/O, or any other layer. Reads
-  `shellContext.__varietyFormatters` to dispatch output. This is the
-  future `@variety/core` package boundary.
+- `core/engine.js` — pure, side-effect-free analysis logic. No runtime
+  dependencies on shell globals, Node I/O, persistence, or formatting. Exports
+  the reducer/finalization engine used by both the shell-facing analyzer and
+  Node-side tests. This is the future `@variety/core` package boundary.
+- `core/analyzer.js` — shell-adjacent orchestration that traverses the cursor,
+  optionally persists results, dispatches to formatters, and exposes `run()`
+  on top of `core/engine.js`. Depends on `core/formatters/` and `core/engine.js`.
 - `mongo-shell/adapter.js` — the shell-facing layer that reads shell
   globals (`collection`, `plugins`, `secondaryOk`, etc.), loads plugins, and
   hands dependencies to `impl.run()`. The only place in the build that
-  touches `db`, `print`, and `load`. Cleans up both `__varietyImpl` and
-  `__varietyFormatters` after execution so repeated loads are idempotent.
+  touches `db`, `print`, and `load`. Cleans up `__varietyEngine`,
+  `__varietyImpl`, and `__varietyFormatters` after execution so repeated loads
+  are idempotent.
   Depends on `core`; compiled into `variety.js` by `build.js` (not a
   separately published package).
 - `bin/variety` — the published Node entrypoint that implements the main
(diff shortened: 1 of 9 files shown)


JamesCropcho:
Overall I think this is a sane, well-scoped seam to cut, and a sensible precondition for #263/#264. A few observations, in rough order of how much I'd act on them.

**Things I'd likely act on**

- JSDoc typos in the test files:
  - [`test/cases/analysis/BinarySubtypeTest.js`](https://github.com/variety/variety/blob/codex/extract-analysis-engine/test/cases/analysis/BinarySubtypeTest.js) and [`test/cases/analysis/DeprecatedBsonTypesTest.js`](https://github.com/variety/variety/blob/codex/extract-analysis-engine/test/cases/analysis/DeprecatedBsonTypesTest.js): on the `varietyTypeOf` line inside the `VarietyImpl` typedef, the leading `*` is indented two extra spaces and no longer aligns with the rest of the block comment. Looks like a stray edit artifact.
  - [`test/cases/analysis/AnalyzerPropertyFuzzingTest.js`](https://github.com/variety/variety/blob/codex/extract-analysis-engine/test/cases/analysis/AnalyzerPropertyFuzzingTest.js): the new `analyzeDocuments` entry in the typedef is missing the trailing comma after `VarietyResultRow[]` on the preceding `finalizeResults` line.

- Dead-on-arrival names on the new engine surface:
  - `reduceDocuments: ingestDocument` is added as a "back-compat alias", but nothing in the tree references the old name after this PR, and `__varietyImpl` (not `__varietyEngine`) was the only previously-registered global, so there is no external consumer to be compatible with. I'd drop the alias — keeping it locks in a name this PR is explicitly trying to move away from.
  - `reduceCursor` is still exported from the engine but no longer called anywhere (the analyzer now iterates the cursor inline). Either remove, or note why it stays as public surface.

- "Pure, side-effect-free" is a slight overstatement for what the engine actually is. `shellToJson` reads `tojson` / `shellContext.EJSON` from the ambient scope, and `getSpecialTypeName` / `varietyTypeOf` reference `NumberLong` / `ObjectId` / `BinData` the same way. The new module is free of *persistence, formatting, and I/O* side effects, which is the real win here — but it still reads shell-provided globals when present. Worth softening the file banner and the CONTRIBUTING blurb so they don't over-promise before a future `@variety/core` extraction has to re-litigate it.

**Things I'd think about but probably leave**

- Public surface of `__varietyEngine`. With this PR, the engine re-exports a fairly wide set (`createKeyMap`, `buildResultFilter`, `compareResults`, `convertResults`, etc.) that are really internal plumbing to `finalizeResults` / `analyzeDocuments`. If this object is going to harden into `@variety/core`, I'd want the externally-documented set to be small (probably `analyzeDocuments`, `createAnalysisState`, `ingestDocument`, `finalizeResults`, `varietyTypeOf`) and treat the rest as internal. Not a blocker for this PR — more a note for whichever follow-up formalizes the package boundary.

- `analyzeDocuments(config, documents, options)` precedence is subtle: it counts by iterating `documents`, then overrides `documentsCount` if the caller passed one. That is the right shape for the "cursor was `.limit()`-ed but `percentContaining` should still be against the full match count" case the shell currently handles, but it's worth one JSDoc line spelling that out so the override doesn't look accidental.

**Things that belong outside this PR (and are correctly outside)**

- The actual Node.js programmatic API (#263) and the interactive shell API (#264). This PR is the seam; those are the next layers. Good call keeping them separate.
- A real `@variety/core` package (separate `package.json`, publish pipeline). This PR deliberately just carves out the module boundary inside the concatenated build — fine as a first step.

**Non-PR note, just for coordination**

- There are uncommitted changes on `main` locally that extend binary-subtype handling (`BinData-user[0x..]`, `BinData-reserved`) in `core/analyzer.js`. Once this PR lands, those edits will need to move into `core/engine.js` (and the corresponding test file hunks re-aimed there). Not this PR's problem, just a merge-order thing to be aware of.

Nothing here looks like a bad idea or a silly mistake beyond the JSDoc nits; the shape of the split reads right.

JamesCropcho:
Following up on the feedback: I pushed `f77b20f`, which addresses the points I agreed with.

- fixed the JSDoc formatting nits in `BinarySubtypeTest.js`, `DeprecatedBsonTypesTest.js`, and `AnalyzerPropertyFuzzingTest.js`
- removed the dead `reduceDocuments` alias and the unused `reduceCursor` export from the new engine surface
- softened the engine-boundary wording in `build.js`, generated `variety.js`, and `CONTRIBUTING.md` so it does not overstate purity
- documented that `analyzeDocuments(..., { documentsCount })` intentionally overrides the iterated count when percentages should be based on a larger matched set

I left the broader `__varietyEngine` surface-pruning note for a later pass rather than expanding the scope of this PR.

JamesCropcho:
`f77b20f` reads clean. Checked it against the review points:

- JSDoc alignment/typedef fixes in all three test files — correct.
- `reduceDocuments` alias and `reduceCursor` export dropped — no remaining references in the tree, and `__varietyImpl`'s shape (via `Object.assign({}, engine, { run })`) just narrows accordingly.
- New engine-boundary wording in [`CONTRIBUTING.md`](https://github.com/variety/variety/blob/codex/extract-analysis-engine/CONTRIBUTING.md), [`build.js`](https://github.com/variety/variety/blob/codex/extract-analysis-engine/build.js), and the generated [`variety.js`](https://github.com/variety/variety/blob/codex/extract-analysis-engine/variety.js) is accurate — "keeps persistence, formatter dispatch, and output side effects out of the engine" is the claim the module actually earns.
- The `analyzeDocuments` JSDoc override note is enough to disambiguate the precedence.

Deferring the `__varietyEngine` surface pruning makes sense — it's a shaping decision that belongs with whichever follow-up actually introduces the external consumer (#263 or #264), since that's what will pin down which helpers need to be part of the public API vs. kept internal.

Nothing else from me on this PR.

PR #335, opened 2026-04-23
Map user-defined binary subtypes 0x80-0xFF to BinData-user[0xNN]; add BinData-reserved for spec-reserved range

JamesCropcho:
## Summary

- Emits `BinData-user[0xNN]` for the full `0x80–0xFF` user-defined subtype range, giving each subtype a distinct label (analogous to the `BinData-vector[dtype]` pattern from #329). Retires the former flat label `BinData-user`.
- Emits `BinData-reserved` for the `0x0A–0x7F` BSON-spec-reserved range (these subtypes cannot be stored in MongoDB; coverage is analyzer-level).
- Retires `BinData-undefined` as a live label for any reachable binary subtype.
- Adds integration fixtures for subtypes `0x80` and `0x81`, plus unit tests for `0xFF` and reserved subtype `0x0A`.
- Updates `docs/bson-type-inventory.md`: adds a reserved-range row, updates the user-defined row to reflect per-subtype labels, and revises the preamble summary.

## Test plan

- [x] Integration test: `binData_user_0x80` → `BinData-user[0x80]` and `binData_user_0x81` → `BinData-user[0x81]` in MongoDB 8.2
- [x] Unit test: subtype `0xff` → `BinData-user[0xff]`
- [x] Unit test: spec-reserved subtype `0x0a` → `BinData-reserved`
- [x] Full container test suite: 89 passing (7 pre-existing flaky failures unrelated to this change)
- [x] All pre-commit hooks pass: build check, ESLint, jsonlint, markdownlint, js-yaml, hadolint, shellcheck, SPDX, typecheck

### DIFF
--- a/core/engine.js
+++ b/core/engine.js
@@ -178,6 +178,12 @@
         if (subtype === 0x09) {
           return getVectorDtypeLabel(thing);
         }
+        if (subtype >= 0x80) {
+          return `BinData-user[0x${subtype.toString(16).padStart(2, '0')}]`;
+        }
+        if (subtype >= 0x0a) {
+          return 'BinData-reserved';
+        }
         const binDataTypes = {
           0x00: 'generic',
           0x01: 'function',
@@ -188,7 +194,6 @@
           0x06: 'encrypted',
           0x07: 'compressed-column',
           0x08: 'sensitive',
-          0x80: 'user',
         };
         return `BinData-${binDataTypes[subtype]}`;
       } else if (typeof specialType !== 'undefined') {
--- a/docs/bson-type-inventory.md
+++ b/docs/bson-type-inventory.md
@@ -50,12 +50,13 @@ type mapping; [V6] tests every standard binary subtype plus custom subtype
 each deprecated top-level BSON type.
 Current tests recognize BSON `Double` and `Int32` as Variety `Number`, BSON
 JavaScript code and JavaScript-with-scope as Variety `Code`, BSON `Symbol`
-(type 14) as Variety `String` in mongosh, standard binary subtypes plus custom
-subtype `0x80` under their `BinData-{subtype}` labels, and vector binary
-subtype `9` values under dtype-specific labels (`BinData-vector[FLOAT32]`,
-`BinData-vector[INT8]`, `BinData-vector[PACKED_BIT]`), with unknown dtypes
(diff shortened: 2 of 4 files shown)


JamesCropcho:
One thing I'd fix before merge:

- `docs/bson-type-inventory.md:48-49` and `docs/bson-type-inventory.md:74` are now out of sync with the detailed subtype table and the implementation. The summary still says V6 covers only custom subtype `0x80`, and the top-level `Binary data` row still describes the custom range as only partially mapped. After this PR, `0x80` through `0xFF` are intentionally mapped to `BinData-user[0xNN]`, `0x0A` through `0x7F` map to `BinData-reserved`, and V6 now covers `0x81`, `0xff`, and `0x0a` as well. The detailed subtype section is right; these summary lines just need to come along so the doc stops contradicting itself.

Otherwise this looks sane to me. It follows the same pattern as #329: take a previously coarse or effectively-undefined label, split it into more semantically accurate output, and back that with targeted tests plus docs. I don't see a code-path mistake in the boundaries here: subtype `0x09` still peels off to vector handling first, `0x0A` is the first reserved value, and `0x80` is the first user-defined value.

Scope also feels right. The behavior change, tests, inventory update, and generated `variety.js` belong together in one PR rather than being split up. The only thing I'd want called out anywhere user-facing is that retiring `BinData-user` is intentionally backward-incompatible for anyone keying off exact Variety labels, but the PR body already makes that pretty clear.

JamesCropcho:
Follow-up complete in `19f75a9`.

I updated the two summary passages in `docs/bson-type-inventory.md` so they now match the detailed subtype table: V6 is described as covering representative reserved/user-defined-range cases, and the top-level binary-data row now reflects the intentional range-wide mappings instead of implying only `0x80` is mapped.

Validation: `npm run lint:markdown`

JamesCropcho:
Correction: my previous status note was premature. The follow-up commit had been created locally as `19f75a9`, but the push step raced it and did not actually send it to GitHub.

That is fixed now: `19f75a9` is on the PR branch, and the PR head has advanced to that commit.

The change itself is still the same narrow doc follow-up:
- update the V6 summary language to describe representative reserved/user-defined-range coverage
- update the top-level binary-data row so it no longer implies only `0x80` is intentionally mapped

Validation remains `npm run lint:markdown`.

Issue #338, opened 2026-04-23
Define a normalized plugin-definition model independent of CLI and shell string syntax


JamesCropcho:
Follow-up from #263 and #264.

## Problem

Variety's current plugin inputs are shaped around shell execution details:

- the CLI accepts repeated `--plugin` flags and serializes them into the shell-facing `plugins` string
- the shell adapter parses comma-delimited plugin definitions with `|` and `&` separators before loading them

That is fine for the current `mongosh ... variety.js` flow, but it is an awkward internal representation for future callable APIs.

## Goal

Define a normalized internal plugin-definition shape that future Node.js and interactive shell APIs can use directly, while preserving today's CLI and shell-compatible string syntax as input adapters.

## Possible direction

- represent plugins internally as structured entries such as `{ path, config }`
- keep the current CLI `--plugin` syntax as one parser / adapter into that normalized form
- keep the current shell `plugins` string compatibility as another parser / adapter
- make future programmatic APIs accept the normalized form directly instead of forcing callers through shell-oriented encoding

## Acceptance sketch

- plugin definitions have one intentional internal shape
- CLI and shell compatibility layers translate into that shape instead of owning the meaning themselves
- plugin loading behavior remains backward compatible
- docs explain both the compatibility syntax and the normalized model

Related: #263 and #264.

Issue #337, opened 2026-04-23
Separate shell transport option handling from shared Variety analysis config


JamesCropcho:
Follow-up from #332 and the upcoming config-centralization work for #263 and #264.

## Problem

`cli/options.js` currently mixes two different concerns:

- Variety analysis options such as `query`, `sort`, `limit`, `maxDepth`, and `outputFormat`
- Mongo shell transport / process-launch options such as `host`, `port`, `username`, `password`, `authenticationDatabase`, and `quiet`

That is workable for the current CLI, but those transport options are launcher concerns. They do not belong in the same reusable config boundary as the analysis settings that future in-process Node.js and interactive shell APIs will need to share.

## Goal

Keep shared analysis config reusable across entrypoints while giving shell transport options their own normalization and validation boundary.

## Possible direction

- keep transport option parsing / normalization in a dedicated launcher-facing module
- continue letting the CLI expose shell transport flags where appropriate
- avoid making future in-process APIs pretend they accept connection-launch settings they do not actually use
- document clearly that transport options are entrypoint concerns, not core analysis config

## Acceptance sketch

- shared Variety analysis config does not need to know about `host`, `port`, or shell authentication flags
- launcher-facing transport options have a dedicated normalization / validation path
- CLI behavior stays backward compatible
- docs explain the boundary between analysis config and shell transport config

Related: #263, #264, #332.

JamesCropcho:
Landed in #351.

- `mongo-shell/transport-options.js` is the dedicated launcher-facing boundary. It owns `ShellOptions`, `SHELL_OPTION_DESCRIPTORS`, and `validateShellOptions` — which rejects unknown keys, type-checks each option via the shared `validateOptions` dispatcher, enforces the `mongodb://` / `mongodb+srv://` URI prefix, and enforces URI vs host/auth-flag exclusivity.
- `mongo-shell/launcher.js` calls `validateShellOptions` inside `buildShellInvocation`, so validation is enforced for any caller, not just the CLI.
- `core/config.js` has no knowledge of `host`, `port`, or auth flags.
- `resolveShellOptions` in `cli/options.js` is now exclusively the target-dependent DB-path injection — every other transport invariant lives in the transport module.
- `CONTRIBUTING.md` documents the boundary and updated dependency directions.

Issue #334, opened 2026-04-23
Tighten the exported surface of `core/engine.js`


JamesCropcho:
Follow-up from #332.

## Problem

`core/engine.js` currently exports a fairly wide surface, including helpers such as `createKeyMap`, `analyseDocument`, `mergeDocument`, `convertResults`, `buildResultFilter`, and `compareResults`.

That is convenient during the extraction step, but it makes the future engine boundary look broader and more public than it probably should be.

## Goal

Shrink the exported surface to the smallest intentional API needed for follow-up work, while keeping implementation helpers private.

## Possible direction

Treat a small set as the intended engine API, likely something close to:

- `analyzeDocuments`
- `createAnalysisState`
- `ingestDocument`
- `finalizeResults`
- `varietyTypeOf`

Everything else should either remain file-local or be explicitly marked as internal implementation detail.

## Acceptance sketch

- `core/engine.js` exports only the intentionally supported engine API
- internal plumbing helpers are no longer exported just because they happen to exist
- tests are updated to target the intended API shape rather than the broader implementation surface
- docs describe the narrower boundary clearly

Related: #263, #264.

JamesCropcho:
Implemented in #352 and now merged into main.

This narrows core/engine.js down to the intentional API (`analyzeDocuments`, `createAnalysisState`, `ingestDocument`, `finalizeResults`, and `varietyTypeOf`), adds an explicit export-surface test so helper re-exports do not creep back in, and documents the boundary in CONTRIBUTING.md.

Closing this out.

PR #339, opened 2026-04-23
Centralize shared analysis config

JamesCropcho:
## Summary
- add shared `core/config.js` for analysis-option validation, default resolution, and engine-facing materialization
- route `cli/options.js` and `mongo-shell/adapter.js` through that shared boundary while keeping `secondaryOk` outside it as a shell runtime option
- update tests, build assembly, generated `variety.js`, and contributor docs for the new config section

## Why
- analysis option semantics were split between the CLI and Mongo shell adapter
- that duplication makes the future callable APIs in #263 and #264 harder to share safely

## Impact
- creates one internal source of truth for analysis-option defaults and validation
- keeps shell transport and plugin-shape follow-ups out of this PR so the shared config boundary stays narrow

## Validation
- `npm run build`
- `npm run verify:build`
- `npm run lint`
- `npm run typecheck`
- `npm run lint:markdown`
- `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/analysis/ConfigTest.js test/cases/cli/CliOptionsTest.js test/cases/mongo-shell/AdapterValidationTest.js`

Refs #263 and #264.
Deferred follow-ups intentionally left out of scope: #334, #337, #338.


JamesCropcho:
Overall take: this is a reasonable, well-scoped refactor. Collapsing two parallel implementations of "what are the Variety analysis options, what do they default to, and what must the engine see" into a single module is the right precursor to #263/#264, and the three-stage split (`validate` → `resolve` → `materialize`) is a clear internal boundary. Keeping `secondaryOk` outside — on the shell-runtime side — is consistent with the deferred #337 split. Scope looks right: the explicit deferrals of #334, #337, and #338 keep this PR narrow and reviewable.

A handful of gestalt observations:

**Log output ordering has shifted.** Pre-PR, the `Using X of Y` echo lines in the shell adapter fired in this order: `collection, query, limit, maxDepth, sort, outputFormat, persistResults, resultsDatabase, resultsCollection, resultsUser, resultsPass, logKeysContinuously, excludeSubkeys, arrayEscape, showArrayElements, compactArrayTypes, lastValue, maxExamples`. Post-PR they follow `ANALYSIS_OPTION_NAMES`: `collection, query, sort, limit, maxDepth, outputFormat, maxExamples, lastValue, showArrayElements, compactArrayTypes, arrayEscape, excludeSubkeys, logKeysContinuously, persistResults, resultsDatabase, resultsCollection, resultsUser, resultsPass`. The PR body says "no intended user-facing behavior change," and practically speaking this is fine — nobody should be parsing those lines — but it is technically a visible diff in `--quiet`-less runs and worth a one-line acknowledgement either in the PR description or in a follow-up, since CI comparisons / screenshots / docs that happen to capture the echo will drift.

**`materializeAnalysisConfig` re-runs `validateAnalysisOptions`.** The contract of `resolveAnalysisOptions` is "returns a fully-populated `ResolvedAnalysisOptions`," so the re-validate + 17-way `hasOwn` check in `materializeAnalysisConfig` is belt-and-suspenders — it re-clones `query` / `sort` / `excludeSubkeys` and then immediately re-checks presence of each key. Not wrong, but it muddies the mental model of what each stage is actually responsible for. Either collapse the two stages on this path (materialize trusts resolved input), or make `materialize` accept the looser `AnalysisOptionsInput` only via a separate entry point so the common "resolve then materialize" path doesn't pay for validation twice. Minor — happy to see it left as-is if you prefer the defensive posture.

**Duplicated typedef in `cli/options.js`.** `AnalysisOptionsInput` is declared in both [core/config.js:13-32](core/config.js#L13-L32) and [cli/options.js:24-43](cli/options.js#L24-L43). Now that `cli/options.js` already requires `core/config.js`, the CLI-side copy can become `/** @typedef {import('../core/config.js').AnalysisOptionsInput} AnalysisOptionsInput */`. Not urgent, but two copies of a 17-field type are going to drift.

**Test coverage is adequate but thin on the error surface.** `ConfigTest.js` covers the happy paths (defaults, explicit overrides, a few validation errors) and the "context required" case. It doesn't exercise: the `resultsUser: null` / `resultsPass: null` allowance, string-type errors on `outputFormat` / `arrayEscape` / `resultsDatabase` / `resultsCollection`, or boolean-type errors on `lastValue` / `showArrayElements` / etc. Most of that surface was already covered end-to-end through the CLI tests before this PR, so nothing is regressing, but the new module is now the canonical home for those errors and its direct tests should eventually be the first line of defense rather than the CLI suite. Fine to leave for a follow-up.

**One nicety for later, not this PR.** The validation function itself is 17 nearly-identical `if (hasOwn(...) && typeof ... !== 'undefined')` blocks. A small table — `{ name, kind: 'nonNegativeInteger' | 'boolean' | 'string' | 'object' | 'stringArray', allowNull?, requireNonEmpty? }` — would make the option list cheap to extend when #337 pulls shell transport out and gives you a second, similarly-shaped validator. Not a refactor to bundle here; flagging only because #337 is explicitly on deck.

Nothing in here looks like a silly mistake or a bad idea, and nothing stood out as wanting to be outside this PR. Ship it once you're comfortable with the log-ordering note.

JamesCropcho:
I applied the feedback items that felt worth landing in this PR:

- Restored the legacy analysis-option ordering so the shell `Using ...` lines and CLI-generated `--eval` assignments keep their pre-refactor order.
- Removed the duplicate CLI-side `AnalysisOptionsInput` declaration by deriving that type from the exported `validateAnalysisOptions()` signature instead.
- Simplified `materializeAnalysisConfig()` so the normal `resolve -> materialize` path no longer re-validates every field; it now treats resolved input as trusted, keeps the required-key guard, and still clones `query` / `sort` for the materialized config.
- Expanded the direct `core/config.js` tests to cover `resultsUser/resultsPass: null`, representative string/boolean type errors, and the “materialize requires resolved input” guard.

I did not pull in the later table-driven validator refactor suggestion here, since I agree that it belongs in a future follow-up rather than broadening this PR.

Checks rerun after the changes:
- `npm run build`
- `npm run verify:build`
- `npm run lint`
- `npm run typecheck`
- `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/analysis/ConfigTest.js test/cases/cli/CliOptionsTest.js test/cases/cli/BinWrapperTest.js test/cases/mongo-shell/AdapterValidationTest.js`

JamesCropcho:
Thanks, the follow-up commit (eb3a70b) addresses every point cleanly and I think this is ready to ship.

Spot checks:

- **Ordering** — `ANALYSIS_OPTION_NAMES` in both `core/config.js` and the committed `variety.js` now reads `query, limit, maxDepth, sort, outputFormat, persistResults, resultsDatabase, resultsCollection, resultsUser, resultsPass, logKeysContinuously, excludeSubkeys, arrayEscape, showArrayElements, compactArrayTypes, lastValue, maxExamples`, which matches the pre-refactor adapter echo order exactly. `CliOptionsTest.js` and `BinWrapperTest.js` assertions were updated in the same direction, so the CLI-generated `--eval` string is pinned to that same order too.
- **Trust boundary on `materializeAnalysisConfig`** — the new `ensureResolvedAnalysisOptions` is a presence guard over `ANALYSIS_OPTION_NAMES`, not a re-validation, and the 17-way OR-chain is gone. The parameter type tightens to `ResolvedAnalysisOptions`, and the defensive `cloneObject(resolved.query)` / `cloneObject(resolved.sort)` keep the materialized config from aliasing caller input — good call keeping those clones.
- **Typedef dedup** — `/** @typedef {NonNullable<Parameters<typeof validateAnalysisOptions>[0]>} AnalysisOptionsInput */` is a nice way to derive the CLI-side type from the config surface without copying fields. Any future field addition in `core/config.js` flows into `cli/options.js` for free.
- **Error-surface tests** — `ConfigTest.js` now covers `resultsUser: null` / `resultsPass: null` acceptance, string-type errors on the four string-valued options, boolean-type errors on a representative three, and the materialize-requires-resolved guard. That's the right sampling — no need to exhaustively enumerate every option.

Declining the table-driven validator for this PR is the right call; I've already opened #343 so it can be the first move when #337 adds the second validator shape. LGTM.

Issue #131, opened 2016-12-14
running from within the mongo shell?


abrin:
Pardon the basic question, but is there a way to load and run variety.js from within a mongo shell?  If so, could the documentation be updated?  Once you "load" the script, how would you call the function.

Separately, I get a weird error in Mongo 3.2.10:
```
Variety: A MongoDB Schema Analyzer
Version 1.5.0, released 14 May 2015
2016-12-14T09:09:19.636-0700 E QUERY    [thread1] Unknown Failure from JSInterpreter
2016-12-14T09:09:19.636-0700 E QUERY    [thread1] Error: error loading js file: /Users/abrin/Documents/workspace/nabo-prototype/src/main/resources/mongo/variety.js :
```

JamesCropcho:
Thanks for getting in touch, Adam. And thank you for using Variety.

Regarding "[s]eparately," kindly create one precise ticket for each important item you're bringing to our attention.

I don't believe there is any way to load Variety into the MongoDB shell, no, if I'm understanding correctly. I mean, other than copy-pasting, but that's not what you likely prefer.

Tomáš / Eve ( @todvora @freeeve ), any thoughts?

todvora:
Hi @abrin ,
This is related to https://github.com/variety/variety/issues/21 and https://github.com/variety/variety/issues/102. There is currently no support for this usage. You would need to wrap the whole function and provide all config attributes from outside. You can get some inspiration on https://github.com/gabrielelana/mongodb-shell-extensions if you decide to do it yourself. Please share with us your experience if you do so.

Regarding the error - is there any line number, expression or anything that could help to locate the exact cause please?

Thanks,
Tomas

abrin:
@todvora thanks.  I don't have any context for the error, unfortunately. I think because of the namespaced / encapsulated functions it simply reports the following:

```
> load("/Users/abrin/Documents/workspace/nabo-prototype/src/main/resources/mongo/variety.js");
Variety: A MongoDB Schema Analyzer
Version 1.5.0, released 14 May 2015
2016-12-19T13:18:53.113-0700 E QUERY    [thread1] Unknown Failure from JSInterpreter
2016-12-19T13:18:53.113-0700 E QUERY    [thread1] Error: error loading js file: /Users/abrin/Documents/workspace/nabo-prototype/src/main/resources/mongo/variety.js :
@(shell):1:1
```

todvora:
Hi @abrin,
[...]

PR #346, opened 2026-04-24
Refactor shell adapter preparation flow for #264

JamesCropcho:
## Summary
- split the shell adapter into explicit preparation and execution steps inside `mongo-shell/adapter.js`
- keep `variety.js` execute-on-load behavior unchanged while moving shell-state gathering behind the new internal seam
- add VM coverage for plugin lifecycle ordering and post-run cleanup of shell handoff globals

## Why
- build toward #264 without jumping ahead to a public interactive shell API
- keep the shell boundary easier to reuse for the later Node.js work in #263 and the planned parity follow-up
- preserve the current shell-specific plugin and transport behavior from #339 rather than folding in #337 or #338

## Impact
- `variety.js` still executes immediately when loaded
- creates a two-step internal shell flow that later work can reuse without making it public yet

## Validation
- `npm run build`
- `npm run verify:build`
- `npm run lint`
- `npm run typecheck`
- `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/mongo-shell/AdapterValidationTest.js`
- focused container-backed smoke test:
  - MongoDB 8.2 / Node.js 22 image built with Podman
  - `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/mongo-shell/SecondaryReadsTest.js` passed inside the container

Refs #264. Original shell-loading question from @abrin: #131.

### DIFF
--- a/mongo-shell/adapter.js
+++ b/mongo-shell/adapter.js
@@ -23,90 +23,105 @@
     resolveAnalysisOptions,
   } = configApi;
 
-  const shellIsQuiet = () => {
-    if (typeof __quiet !== 'undefined' && __quiet) {
+  const shellIsQuiet = (context) => {
+    if (typeof context.__quiet !== 'undefined' && context.__quiet) {
       return true;
     }
 
-    return typeof process !== 'undefined' &&
-      process &&
-      process.argv &&
-      process.argv.includes('--quiet');
+    return typeof context.process !== 'undefined' &&
+      context.process &&
+      context.process.argv &&
+      context.process.argv.includes('--quiet');
   };
 
-  const log = (message) => {
-    if (!shellIsQuiet()) {
-      print(message);
-    }
-  };
-
-  const countMatchingDocuments = (collectionName, query, limit) => {
-    const coll = db.getCollection(collectionName);
-    const options = (typeof limit === 'number' && limit > 0) ? {limit} : undefined;
-    return coll.countDocuments(query, options);
-  };
-
-  log('Variety: A MongoDB Schema Analyzer');
-  log('Version 1.5.2, released 30 September 2025');
-
-  if (typeof secondaryOk !== 'undefined') {
-    if (secondaryOk === true) {
-      db.getMongo().setReadPref('secondary');
-    }
-  }
-
-  const selectedDatabaseName = db.getName();
-  const knownDatabases = db.adminCommand('listDatabases').databases;
-  const knownDatabaseNames = [];
-  if (typeof knownDatabases !== 'undefined') { // not authorized user receives error response (json) without databases key
-    // Keep validation scoped to the selected database. Issue #145
-    // (@pkgajulapalli) hit a startup failure while enumerating collections for
-    // an unrelated database.
-    knownDatabases.forEach((database) => {
-      if (typeof database.name === 'string' && database.name.length > 0) {
-        knownDatabaseNames.push(database.name);
+  const createLogger = (context) => {
+    return (message) => {
+      if (!shellIsQuiet(context)) {
+        context.print(message);
       }
-    });
+    };
+  };
 
-    if (!knownDatabaseNames.includes(selectedDatabaseName)) {
-      throw new Error(`The database specified (${selectedDatabaseName}) does not exist.\n` +
+  const createCountMatchingDocuments = (context) => {
+    return (collectionName, query, limit) => {
+      const coll = context.db.getCollection(collectionName);
+      const options = (typeof limit === 'number' && limit > 0) ? {limit} : undefined;
+      return coll.countDocuments(query, options);
+    };
+  };
+
+  const logBanner = (log) => {
+    log('Variety: A MongoDB Schema Analyzer');
+    log('Version 1.5.2, released 30 September 2025');
+  };
+
+  const applySecondaryReadPreference = (context) => {
+    if (typeof context.secondaryOk !== 'undefined') {
+      if (context.secondaryOk === true) {
+        context.db.getMongo().setReadPref('secondary');
+      }
+    }
+  };
+
+  const validateShellStartup = (context, countMatchingDocuments) => {
+    const selectedDatabaseName = context.db.getName();
+    const knownDatabases = context.db.adminCommand('listDatabases').databases;
+    const knownDatabaseNames = [];
+    if (typeof knownDatabases !== 'undefined') { // not authorized user receives error response (json) without databases key
+      // Keep validation scoped to the selected database. Issue #145
+      // (@pkgajulapalli) hit a startup failure while enumerating collections for
+      // an unrelated database.
+      knownDatabases.forEach((database) => {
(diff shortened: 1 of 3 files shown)


JamesCropcho:
Overall read: sane and low-risk. It's a pure mechanical lift from shell-global reads to context-threaded parameters, the observable behavior is unchanged, and the two new VM tests (plugin lifecycle ordering, handoff-global cleanup) are worth having regardless of the #264 trajectory. No blockers from me.

A few things that nagged at me while reading, roughly in descending order of interest:

**1. `prepareShellExecution` isn't really "just preparation".** It fires `pluginsRunner.execute('onConfig', config)` inside itself. For the #264 shape sketched in the issue (`load(...)` registers callable behavior without running analysis), you almost certainly don't want to invoke user plugin `onConfig` hooks at load time — only at analyze time. So the seam that this PR commits to doesn't quite match the seam #264 will want; `onConfig` dispatch will probably need to move into the execute step (or into a third phase). Not a reason to block, but worth flagging that the "prepare" boundary as drawn here may shift again when #264 actually lands. It would be fine to validate the seam with a throwaway spike before building more on top.

**2. `executePreparedShellExecution` is a one-liner with a heavy name.** It's literally `impl.run(preparedExecution.config, preparedExecution.pluginsRunner, preparedExecution.deps);`. Naming the wrapper + destructuring at the call site would read about the same without the extra indirection. I'd inline it until there's actually a second call site. (The "execute prepared execution" phrasing is also a tautology — if you keep a function, maybe `runPrepared` or `executePreparedShellPlan`.)

**3. `typeof process !== 'undefined'` → `typeof context.process !== 'undefined'` is a subtle semantic shift.** The old form resolves via the scope chain; the new form reads a property off `globalThis`. In mongosh (Node-based) and plain Node these line up. In the legacy SpiderMonkey-based `mongo` shell, I'm less sure whether host-injected identifiers like `process`, `db`, `print`, `connect`, `load` are all properties of `globalThis` vs. scope-only bindings. Given the JavaScript-compatibility floor called out in `build.js` (legacy `mongo` shell still in scope), it's worth either (a) confirming with a quick manual run in the legacy shell, or (b) keeping the `typeof <bareword>` fallback for those specific names. The same concern applies to `context.connect`, `context.print`, `context.load`, `context.db`.

**4. `createCountMatchingDocuments(context)` is overkill for what it captures.** It only needs `context.db` — the factory-of-closures pattern adds a layer that doesn't earn its keep here. `createLogger(context)` is borderline but at least has two context reads (`__quiet` + `process` + `print`). Minor.

**5. The "cleans up shell handoff globals" test only exercises the happy path.** If `validateShellStartup` throws (bad database, missing `collection`, empty collection), the `delete shellContext.__varietyConfig` block at the bottom of the IIFE never runs, so handoff globals leak on failed loads. This is pre-existing behavior — not a regression from this PR — but the new test's presence makes it easy to misread as "we guarantee cleanup" when we don't on the error path. Either broaden the test or tighten its name ("…after successful execution" already does that, actually — fine, withdrawn, but worth keeping in mind if #264 makes re-load-after-failure a supported flow).

**6. Scope.** Fine as one PR. No README/CONTRIBUTING update needed since nothing user-facing changed. The bundled `variety.js` regeneration is obviously correct given `build.js`.

**Nit pile:**
- `validateShellStartup` returns `{ collectionName }` only — the destructuring ceremony at the call site feels heavy for a single string. Returning the string directly would read more naturally.
- `resolveShellConfig(context, collectionName, countMatchingDocuments, log)` — 4 params, where `countMatchingDocuments` is itself a closure over context. Could pass just `(context, collectionName, log)` and re-derive, or fold the signature.
- The JSDoc typedefs in the new test file are dense but consistent with the surrounding style, so fine.

TL;DR: ship it if you want, but I'd inline `executePreparedShellExecution`, and I'd be a little cautious about treating `prepareShellExecution`'s boundary as load-bearing for #264 until you've sketched the actual interactive API.

JamesCropcho:
Addressed the feedback I agreed with in `b111dd6`.

Changes made:
- moved `pluginsRunner.execute('onConfig', config)` out of `prepareShellExecution(...)` so the prepare step now stops before plugin hooks run
- inlined the trivial `executePreparedShellExecution(...)` wrapper
- added bareword fallbacks for shell-provided bindings (`process`, `print`, `db`, `connect`, `load`) instead of assuming those always come from `globalThis` properties
- simplified the startup helper to return `collectionName` directly
- simplified the count helper so it captures only the shell DB handle instead of the whole context

Validation rerun after the update:
- `npm run build`
- `npm run verify:build`
- `npm run lint`
- `npm run typecheck`
- `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/mongo-shell/AdapterValidationTest.js`
- focused Podman-backed smoke test: `test/cases/mongo-shell/SecondaryReadsTest.js`

I did not change the withdrawn cleanup-on-error point or add docs, since those still looked out of scope for this PR.

JamesCropcho:
All five look cleanly addressed. Reviewing the revised shape at `b111dd6`:

- **`onConfig` dispatch outside `prepareShellExecution`** — the top-level call sequence (prepare → fire plugin `onConfig` → `impl.run`) with the explanatory comment at [mongo-shell/adapter.js:200-202](mongo-shell/adapter.js#L200-L202) is the right shape for a future callable API. Good.
- **Inlined `impl.run(...)`** — reads fine now.
- **Bareword fallbacks** via `getShellProcess` / `getShellPrint` / `getShellDb` / `getShellConnect` / `getShellLoad` — this is the right hedge against the SpiderMonkey scope-vs-property asymmetry. Worth noting for the record that you (correctly) didn't apply the same fallback to `__quiet`, `secondaryOk`, `collection`, or `plugins` — those are set via user `var` declarations from `--eval`, which *are* reified on `globalThis` in both mongosh and the legacy shell, so `context.x` is sufficient. The asymmetry between host-injected bindings and user `var`s is exactly the right line to draw.
- **`createCountMatchingDocuments(shellDb)`** — the narrower signature now advertises its actual dependency. Nice.
- **`validateShellStartup` returning the string directly** — cleaner.

One small shape observation, purely optional:

- `getShellDb(context)` is now called three times (inside `prepareShellExecution`, `applySecondaryReadPreference`, and `validateShellStartup`). Each call redoes the same `typeof` probe. You could resolve `shellDb` once in `prepareShellExecution` and pass it into the helpers that need it — `applySecondaryReadPreference(shellDb, context)` and `validateShellStartup(shellDb, context, countMatchingDocuments)`. Tiny readability and symmetry win with `createCountMatchingDocuments(shellDb)`. Not worth another round-trip; flagging only because it's the one place where the context-threading still feels a touch heavier than it needs to be.

I also still think `resolveShellConfig(context, collectionName, countMatchingDocuments, log)` could drop `countMatchingDocuments` and inline the closure, but that's the same nit from last round and genuinely a matter of taste.

LGTM to merge.

PR #349, opened 2026-04-25
Table-drive analysis option validation; extract shared dispatcher (#343)

### LINKED ISSUE #343: Table-drive analysis option validation in core/config.js
Follow-up from the review of #339.

## Background

#339 introduced [core/config.js](core/config.js) as the shared boundary for analysis-option validation, default resolution, and engine-facing materialization. Inside that module, `validateAnalysisOptions` is structured as 17 nearly-identical guard blocks, one per option:

```js
if (hasOwn(source, 'query') && typeof source['query'] !== 'undefined') {
  validated.query = validateObjectOption('query', source['query']);
}

if (hasOwn(source, 'sort') && typeof source['sort'] !== 'undefined') {
  validated.sort = validateObjectOption('sort', source['sort']);
}

// ... 15 more of these
```

That shape was fine for the initial extraction — it's explicit and easy to read — but it makes the option list expensive to extend, and the per-option typing knobs (`allowNull`, `requireNonEmpty`) are encoded as positional arguments at each call site rather than as data.

## Relationship to #337

#337 is going to pull shell transport options (`host`, `port`, `username`, `password`, `authenticationDatabase`, `quiet`) out of the shared config boundary and into their own launcher-facing normalization / validation path. That is a second validator with the same shape as the current one: a list of named options, each with a kind and a few optional constraints.

Before that second validator exists, it's worth factoring the shared shape into something data-driven so #337 inherits the table rather than copying another 6–10 if-blocks.

## Goal

Replace the per-option guard blocks in `validateAnalysisOptions` with a single table and a small dispatcher, so both the current analysis-option validator and the forthcoming transport-option validator from #337 can share the same engine.

## Possible direction

- Define a descriptor shape roughly like:
  ```js
  { name, kind: 'nonNegativeInteger' | 'boolean' | 'string' | 'object' | 'stringArray',
    allowNull?: boolean, requireNonEmpty?: boolean }
  ```
- Express `ANALYSIS_OPTION_NAMES` as an array of those descriptors (or keep the names array and add a parallel descriptor map).
- Have `validateAnalysisOptions` iterate the table and dispatch to the existing `validateBooleanOption` / `validateStringOption` / `validateObjectOption` / `validateStringArrayOption` / `validateNonNegativeIntegerOption` helpers by `kind`.
- Keep JSDoc typing tight — the output type should still be the current `AnalysisOptionsInput`-shaped projection, not a loose `Record<string, unknown>`.
- When #337 lands, the transport-option validator should reuse the same dispatcher with its own descriptor table.

## Acceptance sketch

- `validateAnalysisOptions` is data-driven off a descriptor table rather than 17 hand-written guards.
- Adding a new analysis option is a one-line table entry plus a default in `resolveAnalysisOptions`, not a new if-block.
- The transport-option validator introduced by #337 reuses the same dispatch mechanism instead of duplicating it.
- `npm run typecheck` still passes with `exactOptionalPropertyTypes` / `noUncheckedIndexedAccess` / `noPropertyAccessFromIndexSignature` on; typings around the dispatcher stay strict rather than falling back to `any` / `unknown`.
- Existing direct tests in `test/cases/analysis/ConfigTest.js` and CLI-side validation error messages are unchanged.

Related: #263, #264, #337, #339.

JamesCropcho:
## Summary

Closes #343. Follow-up from the #339 review.

- **New `core/option-validation.js`** — `OptionDescriptor` typedef and `validateOptions` dispatcher (plus the five typed validator helpers). Registers as `__varietyOptionValidation` for shell-context use; exports via `module.exports` for Node. No deps on any other layer.
- **`core/config.js`** — `ANALYSIS_OPTION_DESCRIPTORS` table replaces 17 hand-written guard blocks; `ANALYSIS_OPTION_NAMES` is derived from it via `.map()`; `validateAnalysisOptions` collapses to a one-liner. The four validator helpers previously duplicated here are removed (only `validateNonNegativeIntegerOption` stays, used by `resolveDefaultLimit`).
- **`build.js`** — `option-validation.js` slotted before `config.js` in the concatenation; section counts and file-list comments updated.
- **`.tsconfig.checkjs.json`, `.eslint.config.js`** — `option-validation.js` added to type-checking and shell-compat lint targets.
- **`CONTRIBUTING.md`** — repo layout, dependency directions, and typecheck descriptions updated.

Adding a new analysis option is now a one-line table entry in `ANALYSIS_OPTION_DESCRIPTORS` plus a default in `resolveAnalysisOptions`. When #337 lands, the transport-option validator can call `validateOptions` with its own descriptor table instead of copying another batch of if-blocks.

## Test plan

- [x] `npm run typecheck` passes (`exactOptionalPropertyTypes` / `noUncheckedIndexedAccess` / `noPropertyAccessFromIndexSignature` all on)
- [x] `npm run lint` passes (all linters including SPDX, markdownlint, shellcheck, hadolint)
- [x] `npm run verify:build` passes
- [x] All 6 `ConfigTest.js` cases pass (unchanged assertions on `ANALYSIS_OPTION_NAMES` order, `validateAnalysisOptions` error messages, `resolveAnalysisOptions` defaults, and `materializeAnalysisConfig` output)
- [x] 78 total unit tests pass; 19 MongoDB-integration timeouts are pre-existing (no local `mongod`; `npm run test:docker` covers those in CI)

### DIFF
--- a/.eslint.config.js
+++ b/.eslint.config.js
@@ -89,7 +89,7 @@ module.exports = (async () => {
       rules: nodeModernizationRules,
     },
     {
-      files: ['variety.js', 'core/config.js', 'core/engine.js', 'core/analyzer.js', 'mongo-shell/adapter.js'],
+      files: ['variety.js', 'core/option-validation.js', 'core/config.js', 'core/engine.js', 'core/analyzer.js', 'mongo-shell/adapter.js'],
       // Rules safe for the ES6+ compatibility floor declared at the top of variety.js.
       // prefer-object-has-own is excluded because Object.hasOwn() (ES2022) is absent
       // from the legacy mongo shell; see the compatibility comment in variety.js.
--- a/.tsconfig.checkjs.json
+++ b/.tsconfig.checkjs.json
@@ -23,6 +23,7 @@
   "include": [
     "bin/variety",
     "cli/**/*.js",
+    "core/option-validation.js",
     "core/config.js",
     "mongo-shell/launcher.js",
     ".eslint.config.js",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,16 +22,21 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
 
 ## Repo Layout and the `variety.js` Build
 
-`variety.js` at the repo root is a generated file, assembled from six sources in this order:
+`variety.js` at the repo root is a generated file, assembled from seven sources in this order:
 
 - `core/formatters/ascii.js` — built-in ASCII table formatter. Self-contained
   IIFE; registers an `ascii` factory on `shellContext.__varietyFormatters`.
 - `core/formatters/json.js` — built-in JSON formatter. Self-contained IIFE;
   registers a `json` factory on `shellContext.__varietyFormatters`.
+- `core/option-validation.js` — shared, table-driven option validation
+  dispatcher. Registers `validateOptions` on
+  `shellContext.__varietyOptionValidation` for use by `core/config.js` and any
+  future option-validation boundaries (e.g. transport options in #337). Has no
+  deps on any other layer.
 - `core/config.js` — shared analysis-option validation, default resolution,
   and engine-facing materialization. This is the reusable config boundary for
   follow-up work on callable APIs, while shell transport and launch settings
-  remain entrypoint concerns.
+  remain entrypoint concerns. Depends on `core/option-validation.js`.
 - `core/engine.js` — reusable analysis logic that keeps persistence,
   formatter dispatch, and output side effects out of the engine. It still
   tolerates shell/runtime helpers when they are available. Exports the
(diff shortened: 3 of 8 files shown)


JamesCropcho:
I went through this mostly as a design/omission review rather than a green-checks review. Overall, this looks sane to me: the table-driven shape is a reasonable cleanup of the 17 repeated guard blocks in `core/config.js`, and it lines up well with the follow-on `#337` work instead of feeling like abstraction for its own sake. I also think the `build.js` / `CONTRIBUTING.md` updates belong here rather than in a separate PR, because the new file becomes part of both the generated artifact and the contributor-facing repo layout.

One thing I would fix before merging:

- `core/option-validation.js:169` now registers a new shell-global handoff at `__varietyOptionValidation`, but the cleanup block in `mongo-shell/adapter.js:203-208` still only deletes `__varietyConfig`, `__varietyEngine`, `__varietyImpl`, and `__varietyFormatters`. After this PR, loading `variety.js` will leave the new internal handoff behind on `globalThis`, which makes the adjacent “no ad hoc internals leak” comment inaccurate and weakens the repeated-load hygiene that block is trying to preserve. I think this PR should add `delete shellContext.__varietyOptionValidation;` there.

A couple of non-blocking observations:

- The abstraction itself seems reasonable, but the type boundary is still basically a runtime helper plus a cast in `core/config.js:205-207` (`validateOptions(...)` returns `Record<string, unknown>`, then `validateAnalysisOptions` casts it back to `AnalysisOptionsInput`). I would not block this PR on that, but I also would not want to claim we now have a strongly typed shared validator API yet; that is still something to keep an eye on when `#337` adds the second consumer.
- I do not think this needs to be split further. The docs/build changes feel properly in-scope. If anything is a candidate for a later follow-up, it would be giving `core/option-validation.js` its own focused test file once it has more than one caller, but I would not hold this PR for that today.

JamesCropcho:
Thanks for the review.

**Applied:** Added `delete shellContext.__varietyOptionValidation;` to the cleanup block in `mongo-shell/adapter.js` (new commit `b135353`), ahead of the existing four deletions. All pre-commit hooks pass on the new commit.

**Not acted on:** The cast observation (`validateOptions` returning `Record<string, unknown>` cast back to `AnalysisOptionsInput`) is noted — agreed it's the right place to keep an eye on that boundary once #337 adds a second consumer, but not something to resolve here.

PR #351, opened 2026-04-27
Separate shell transport options into `mongo-shell/transport-options.js` (#337)

### LINKED ISSUE #337: Separate shell transport option handling from shared Variety analysis config

JamesCropcho:
Closes #337.

## Summary

- **`core/option-validation.js`** — adds `positiveInteger` kind to the `OptionKind` union and a `validatePositiveIntegerOption` helper, so port validation fits the same table-driven pattern as every other option kind.
- **`mongo-shell/transport-options.js`** (new) — dedicated normalization and validation boundary for shell transport options (`host`, `port`, `username`, `password`, `authenticationDatabase`, `quiet`, `uri`). Declares `ShellOptions` as the canonical typedef, defines `SHELL_OPTION_DESCRIPTORS`, and exposes `validateShellOptions`, which:
  - rejects unknown keys (programmatic callers cannot silently misspell a transport option),
  - delegates type checks to the shared `validateOptions` dispatcher,
  - enforces the `mongodb://` / `mongodb+srv://` URI prefix,
  - enforces URI vs host/auth-flag exclusivity (`uri` cannot be combined with `host`, `port`, `username`, `password`, or `authenticationDatabase`).
  Node-only; not compiled into `variety.js`.
- **`mongo-shell/launcher.js`** — removes the now-redundant local `ShellOptions` typedef (replaced by a `@typedef` import); calls `validateShellOptions` at the top of `buildShellInvocation` so any non-CLI caller going through the standard launch path gets canonical validation automatically.
- **`cli/options.js`** — removes the local inline `ShellOptions` typedef and `parseMongoUri` (URI-shape validation is now exclusively owned by the transport module); imports `validateShellOptions` from `mongo-shell/transport-options.js` and calls it at the end of `parseCliArguments`; `resolveShellOptions` is stripped to only the target-dependent DB-path injection — the one piece that genuinely belongs in CLI land.
- **`.tsconfig.checkjs.json`**, **`.eslint.config.js`** — `transport-options.js` added to typecheck targets and the Node modernization lint set.
- **`CONTRIBUTING.md`** — repo layout, dependency directions, and typecheck description updated.

## Design

The transport module sits in `mongo-shell/` rather than `cli/` because transport options are launcher concerns, not CLI concerns. The dependency arrow is: `cli/options.js` → `mongo-shell/transport-options.js` → `core/option-validation.js`; `mongo-shell/launcher.js` → `mongo-shell/transport-options.js`. `resolveShellOptions` stays in `cli/options.js` because DB-path injection requires a resolved `target` (database name) and is therefore inherently CLI-bound. Every other transport invariant lives in the transport module.

`variety.js` and `build.js` are unchanged apart from the `positiveInteger` kind added to the shared `option-validation.js` IIFE (which was already part of the build).

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm run lint` passes (all linters including SPDX, markdownlint, shellcheck, hadolint)
- [x] `npm run verify:build` passes
- [x] 78 unit tests pass; 19 MongoDB-integration timeouts are pre-existing (`npm run test:docker` covers those in CI)

### DIFF
--- a/.eslint.config.js
+++ b/.eslint.config.js
@@ -84,7 +84,7 @@ module.exports = (async () => {
       },
     },
     {
-      files: ['.eslint.config.js', 'bin/variety', 'build.js', 'cli/**/*.js', 'mongo-shell/launcher.js', 'test/**/*.js'],
+      files: ['.eslint.config.js', 'bin/variety', 'build.js', 'cli/**/*.js', 'mongo-shell/transport-options.js', 'mongo-shell/launcher.js', 'test/**/*.js'],
       ignores: ['test/fixtures/**/*.js'],
       rules: nodeModernizationRules,
     },
--- a/.tsconfig.checkjs.json
+++ b/.tsconfig.checkjs.json
@@ -25,6 +25,7 @@
     "cli/**/*.js",
     "core/option-validation.js",
     "core/config.js",
+    "mongo-shell/transport-options.js",
     "mongo-shell/launcher.js",
     ".eslint.config.js",
     "build.js",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,7 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
 - `core/option-validation.js` — shared, table-driven option validation
   dispatcher. Registers `validateOptions` on
   `shellContext.__varietyOptionValidation` for use by `core/config.js` and any
-  future option-validation boundaries (e.g. transport options in #337). Has no
-  deps on any other layer.
+  other option-validation boundaries. Has no deps on any other layer.
 - `core/config.js` — shared analysis-option validation, default resolution,
   and engine-facing materialization. This is the reusable config boundary for
   follow-up work on callable APIs, while shell transport and launch settings
@@ -55,6 +54,12 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
   separately published package).
 - `bin/variety` — the published Node entrypoint that implements the main
   CLI surface.
+- `mongo-shell/transport-options.js` — dedicated normalization and validation
+  boundary for shell transport options (`host`, `port`, `username`, `password`,
+  `authenticationDatabase`, `quiet`, `uri`). Reuses the `validateOptions`
+  dispatcher from `core/option-validation.js` with its own descriptor table.
+  The canonical definition of `ShellOptions`. Node-only; not compiled into
+  `variety.js`.
 - `mongo-shell/launcher.js` — Node-side Mongo shell invocation helpers.
   The only place that touches Node `spawnSync` for spawning `mongosh`/`mongo`.
   Future interfaces (MCP, etc.) can import from `mongo-shell/` without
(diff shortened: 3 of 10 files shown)


JamesCropcho:
A couple of design concerns before I'd feel great about this landing:

1. `mongo-shell/transport-options.js` is presented as the launcher-facing transport boundary, but the only runtime use in this PR is the new `validateShellOptions()` call at the end of `parseCliArguments()` in `cli/options.js`. `mongo-shell/launcher.js` still accepts arbitrary `shellOptions` directly. That means future non-CLI callers still will not get the canonical validation path unless they remember to opt into it themselves, which feels a bit at odds with #337's goal of making transport options a launcher concern rather than a CLI concern.

2. The new module centralizes the shape/type checks, but the more meaningful transport invariants are still split across `cli/options.js`: `parseMongoUri()` still owns the `mongodb://` / `mongodb+srv://` rule, and `resolveShellOptions()` still owns the `--uri` vs `--host` / `--port` / auth-flag exclusivity plus DB/path reconciliation. Because of that, this PR adds a new shared `positiveInteger` kind and a new module, but the interesting transport behavior is still not actually centered in the transport module. My instinct would be either:
   - move those transport-specific rules into `mongo-shell/transport-options.js` too, leaving only the target-dependent DB injection/reconciliation in CLI land, or
   - scope this PR down more explicitly as typedef/descriptor extraction, and leave the full boundary move to follow-up work.

Open question: if `validateShellOptions()` is meant to become the programmatic boundary for future callers, do we really want it to inherit `validateOptions()`'s silent-ignore behavior for unknown keys? That is fine for CLI-built objects, but it could mask misspelled transport keys for direct consumers.

Directionally I like extracting the canonical `ShellOptions` typedef and documenting the boundary more clearly. I just think this current cut stops half a step before the real responsibility move, so I'd be a little nervous about merging it as though #337 were fully wrapped up.

JamesCropcho:
Agreed with all three points; pushed a follow-up commit (e98ede4) that addresses each.

**1. Validation enforced at the launcher boundary.**
`buildShellInvocation` in `mongo-shell/launcher.js` now calls `validateShellOptions` at the top, so any non-CLI caller going through the standard invocation path gets canonical validation without having to opt in manually.

**2. Transport-specific rules moved into the transport module.**
`validateShellOptions` now enforces both the URI format rule (`mongodb://` / `mongodb+srv://` prefix) and the URI vs host/auth-flag exclusivity rule. `resolveShellOptions` in `cli/options.js` is left with only the target-dependent DB-path injection — the one piece that genuinely belongs in CLI land.

**3. Unknown-key rejection added.**
`validateShellOptions` now iterates `Object.keys(input)` against `KNOWN_SHELL_OPTION_KEYS` and throws for any unrecognized name before delegating to `validateOptions`. Programmatic callers with a misspelled transport key will get an error rather than a silent no-op.

One knock-on effect: the conflict error message changes from `--uri cannot be combined with --host` (CLI prefix) to `uri cannot be combined with host` (transport module, no CLI assumption). Test assertions in `BinWrapperTest.js` and `CliOptionsTest.js` updated to match.

JamesCropcho:
Re-reviewing the current head after `e98ede4`: this is much better, and the main architectural concern I had is addressed now.

Validating at `buildShellInvocation()` closes the non-CLI hole, moving the URI/conflict invariants into `validateShellOptions()` makes the transport boundary real instead of just documentary, and rejecting unknown keys is the right contract for programmatic callers.

One small residual nit: `cli/options.js` still has `parseMongoUri()` enforcing the `mongodb://` / `mongodb+srv://` prefix before `validateShellOptions()` runs, so that particular rule now exists in both the CLI parser and the transport module. I don’t think that is worth holding this PR for, but if you want the separation to be maximally crisp, that helper could shrink to “non-empty string parsing” and let the transport module be the only owner of URI-shape validation.

With that caveat, I no longer have the earlier “#337 is only half done” concern.

JamesCropcho:
Agreed, and went ahead with it in this PR (7908f42).

`parseMongoUri()` and its local `MONGODB_URI_PREFIXES` constant are removed from `cli/options.js`. The one call site in the `uri` switch case is replaced with `parseNonEmptyString()`. The transport module is now the sole owner of the URI-shape rule.

PR #352, opened 2026-04-27
Tighten the exported surface of core/engine.js (#334)

### LINKED ISSUE #334: Tighten the exported surface of `core/engine.js`

JamesCropcho:
Closes #334.

## Summary
- narrow `core/engine.js` exports to the intentional engine API: `analyzeDocuments`, `createAnalysisState`, `ingestDocument`, `finalizeResults`, and `varietyTypeOf`
- add an explicit analysis test that locks the engine export surface so helper re-exports do not creep back in
- document the intentional engine boundary in `CONTRIBUTING.md` and keep the shell-facing `impl.shellToJson` path working from `core/analyzer.js`

## Why
- `core/engine.js` was still exporting a broad extraction-era surface that made the future engine boundary look more public than intended
- this keeps implementation helpers file-local while preserving the reducer/finalization API used by the analyzer and engine-focused tests

## Validation
- `npm run build`
- `node build.js --check`
- `/home/j/variety/node_modules/.bin/tsc --project .tsconfig.checkjs.json`
- `/home/j/variety/node_modules/.bin/eslint --config .eslint.config.js core/engine.js core/analyzer.js test/cases/analysis/EngineExportSurfaceTest.js`
- `npm run test:container`
  - the container lane reached `125 passing` before four late-suite Mongo connectivity/timeouts in `ContinuousLoggingTest`, `PersistenceTest`, and `PluginTest` (`MongoServerSelectionError: connect ECONNREFUSED 127.0.0.1:27017`)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,9 +41,10 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
   tolerates shell/runtime helpers when they are available. Intentionally
   exports only `analyzeDocuments`, `createAnalysisState`, `ingestDocument`,
   `finalizeResults`, and `varietyTypeOf`; implementation helpers stay
-  file-local. This reducer/finalization engine is used by both the
-  shell-facing analyzer and Node-side tests, and is the future `@variety/core`
-  package boundary.
+  file-local, while `varietyTypeOf` remains public as the direct
+  type-classification entrypoint used by focused tests and future callers.
+  This reducer/finalization engine is used by both the shell-facing analyzer
+  and Node-side tests, and is the future `@variety/core` package boundary.
 - `core/analyzer.js` — shell-adjacent orchestration that traverses the cursor,
   optionally persists results, dispatches to formatters, and exposes `run()`
   on top of `core/engine.js`. Depends on `core/formatters/` and `core/engine.js`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,9 +38,12 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
   remain entrypoint concerns. Depends on `core/option-validation.js`.
 - `core/engine.js` — reusable analysis logic that keeps persistence,
   formatter dispatch, and output side effects out of the engine. It still
-  tolerates shell/runtime helpers when they are available. Exports the
-  reducer/finalization engine used by both the shell-facing analyzer and
-  Node-side tests. This is the future `@variety/core` package boundary.
+  tolerates shell/runtime helpers when they are available. Intentionally
+  exports only `analyzeDocuments`, `createAnalysisState`, `ingestDocument`,
+  `finalizeResults`, and `varietyTypeOf`; implementation helpers stay
+  file-local. This reducer/finalization engine is used by both the
+  shell-facing analyzer and Node-side tests, and is the future `@variety/core`
+  package boundary.
 - `core/analyzer.js` — shell-adjacent orchestration that traverses the cursor,
   optionally persists results, dispatches to formatters, and exposes `run()`
   on top of `core/engine.js`. Depends on `core/formatters/` and `core/engine.js`.
--- a/core/analyzer.js
+++ b/core/analyzer.js
@@ -16,6 +16,18 @@
     throw new Error('Expected core/engine.js to register __varietyEngine.');
   }
 
+  const shellToJson = (value) => {
+    if (typeof tojson === 'function') {
+      return tojson(value);
+    }
+
+    if (shellContext.EJSON && typeof shellContext.EJSON.stringify === 'function') {
+      return shellContext.EJSON.stringify(value);
+    }
+
+    return JSON.stringify(value);
+  };
+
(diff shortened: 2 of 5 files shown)


JamesCropcho:
**Overall:** Sane, clean, and the stated goal is accomplished. A few observations below — none are blockers.

---

**`shellToJson` is now defined twice**

The function lives file-locally in `core/engine.js` (used by `getBinDataHex`'s fallback branch) and again in `core/analyzer.js`. The duplication is intentional — `getBinDataHex` needs it internally but shouldn't surface it, while `adapter.js` needs it on `impl` — but it's worth noting explicitly. If `getBinDataHex`'s fallback branch ever changes, both copies would need updating. A follow-on comment in `engine.js` marking it as a local copy would prevent future confusion.

---

**Is `varietyTypeOf` intentionally in the public engine API?**

The five exported functions split naturally into two groups:
- *Lifecycle/reducer*: `createAnalysisState`, `ingestDocument`, `finalizeResults`, `analyzeDocuments`
- *Type classification*: `varietyTypeOf`

`varietyTypeOf` appears to be exported so that `BinarySubtypeTest.js` and `DeprecatedBsonTypesTest.js` can call it directly (`impl.varietyTypeOf(...)`). If the only reason it's in the API is testability, that's worth knowing — either document the intent ("exported for direct testing of the classification logic") or consider whether those tests could call through `analyzeDocuments` instead. Not a problem either way, just naming it.

---

**`impl` still spreads all five engine exports**

`Object.assign({}, engine, { run, shellToJson })` puts all engine functions on `impl`. `adapter.js` only ever calls `impl.run()` and `impl.shellToJson()` — the engine functions riding along via the spread are not needed by the adapter. The spread is probably intentional (shell users could access them via `impl` in the REPL, or Node-side tests that load `analyzer.js` may rely on the passthrough), but it means the `impl` surface is still wide even though the `engine` surface is now tight. This PR doesn't make it worse, but the tension is worth tracking.

---

**`EngineExportSurfaceTest.js`**

The snapshot test is the right call here. A few micro-observations:
- The test uses `assert` from `'assert'` not `'assert/strict'` — consistent with the file header but unlike the sibling test files (`BinarySubtypeTest.js`, `DeprecatedBsonTypesTest.js`) which use `'assert/strict'`. Probably fine since `deepEqual` is strict about order for arrays in both modes; just inconsistent.
- Loading `engine.js` via `require()` also sets `globalThis.__varietyEngine = engine` as a side effect (the IIFE runs). Pre-existing pattern in this test directory, not new here, but worth being aware of for test isolation.

---

**`variety.js` sync**

The bundled file is updated in the same commit as the sources, which is correct. The validation section confirms `npm run build` was run. No concerns.

---

**`CONTRIBUTING.md`**

The updated description is accurate and precise. Good.

JamesCropcho:
Follow-up from the feedback comment:

- I added an explicit comment in `core/engine.js` calling out that its `shellToJson` is a local copy for `getBinDataHex()`'s fallback path, while the analyzer keeps its own copy for shell-facing logging. That should make the intentional duplication less surprising.
- I also clarified the `varietyTypeOf` intent in two places:
  - a code comment above `varietyTypeOf()` saying it remains in the engine API so direct callers can exercise classification without a full document-analysis pass
  - `CONTRIBUTING.md`, which now says `varietyTypeOf` stays public as the direct type-classification entrypoint for focused tests and future callers

I did not change the `impl` spread in `core/analyzer.js`; I read that part as a useful design observation to track, but changing it here would have broadened the scope beyond the `core/engine.js` export-tightening in #334.

Also, on the `EngineExportSurfaceTest.js` point: by the time I revisited the branch it was already using `assert/strict`, so there was nothing further to change there.

JamesCropcho:
The additions land well. The `shellToJson` comment in `engine.js` makes the duplication intentional rather than overlooked, and the two-place clarification of `varietyTypeOf` (code comment + CONTRIBUTING) is exactly the right level of documentation for a boundary decision like this. Leaving the `impl` spread out of scope is the right call.

One correction to my earlier note: looking back at the diff, `EngineExportSurfaceTest.js` was already `import assert from 'assert/strict'` from the start — I misread it and the critique shouldn't have been there. Apologies for the noise.

LGTM.

PR #355, opened 2026-04-27
Speed up YAML linting

JamesCropcho:
## Summary
- replace the per-file `xargs -n1 ./node_modules/.bin/js-yaml` pipeline with a one-process `scripts/lint-yaml.js` runner
- feed that runner from `git ls-files -z -- '*.yml' '*.yaml'` so YAML linting stays scoped to git-tracked files instead of a filesystem deny-list
- preserve `js-yaml` parser diagnostics while prefixing failures with the YAML file path and continuing through all files

## Why
The previous `lint:yaml` command started a fresh Node process for every YAML file. In this checkout that also meant linting YAML files from nested worktrees, which made `npm run lint:yaml` much slower than the actual parsing work justified. Using `git ls-files` also keeps the file-selection behavior aligned with other repo-owned tooling like `scripts/check-spdx.js`.

## Scope
This PR intentionally stays scoped to YAML linting. JSON and Markdown still use their existing file-enumeration paths and can be handled separately.

## Impact
- `npm run lint:yaml` now stays within one Node process
- YAML linting now targets git-tracked YAML files
- failure output remains actionable and now includes the filename
- contributor docs now point to the repo-owned YAML lint runner and its tracked-file semantics

## Validation
- `npm run lint:yaml`
- `node /home/j/variety/node_modules/eslint/bin/eslint.js --config .eslint.config.js scripts/lint-yaml.js`
- `npm run lint:spdx`
- `git diff --check`
- targeted failure check in a temp git repo with tracked `good.yml` and `bad.yml`
- local timing check: `npm run lint:yaml` dropped from about `17.8s` to about `0.29s` in this checkout

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -175,7 +175,7 @@ Otherwise `npm run lint:pre-commit` runs all of the following in parallel via [n
 - `npm run lint` — ESLint (JavaScript, including fenced code blocks in Markdown)
 - `npm run lint:json` — `@prantlf/jsonlint` (JSON files)
 - `npm run lint:markdown` — markdownlint (Markdown files)
-- `npm run lint:yaml` — js-yaml (YAML files)
+- `npm run lint:yaml` — `scripts/lint-yaml.js` using `js-yaml` (YAML files)
 - `npm run lint:dockerfile` — hadolint (`docker/Dockerfile.template`); requires `hadolint` on `PATH` (`dnf install hadolint` / `brew install hadolint`)
 - `npm run lint:shell` — shellcheck (shell scripts); requires `shellcheck` on `PATH` (`dnf install ShellCheck` / `brew install shellcheck`)
 - `npm run lint:spdx` — verifies `SPDX-License-Identifier: MIT` headers in all tracked source files
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -175,7 +175,7 @@ Otherwise `npm run lint:pre-commit` runs all of the following in parallel via [n
 - `npm run lint` — ESLint (JavaScript, including fenced code blocks in Markdown)
 - `npm run lint:json` — `@prantlf/jsonlint` (JSON files)
 - `npm run lint:markdown` — markdownlint (Markdown files)
-- `npm run lint:yaml` — `scripts/lint-yaml.js` using `js-yaml` (YAML files)
+- `npm run lint:yaml` — `scripts/lint-yaml.js` using `js-yaml` (git-tracked YAML files)
 - `npm run lint:dockerfile` — hadolint (`docker/Dockerfile.template`); requires `hadolint` on `PATH` (`dnf install hadolint` / `brew install hadolint`)
 - `npm run lint:shell` — shellcheck (shell scripts); requires `shellcheck` on `PATH` (`dnf install ShellCheck` / `brew install shellcheck`)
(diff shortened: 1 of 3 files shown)


JamesCropcho:
**Overall: sane and worth merging.** The direction is right (one Node process instead of N), the output format is preserved with a useful filename prefix added, and 17.8s → 0.25s is plausible given that the old find was descending into `.claude/worktrees/*/.github/workflows/*.yml` etc. The new script is small, readable, and matches the SPDX/header conventions in `scripts/check-spdx.js`.

A few things worth considering — in roughly decreasing order of substance:

1. **Why not `git ls-files`?** `scripts/check-spdx.js` already takes that approach, and it would naturally skip everything in `.gitignore` (`.claude/`, `.codex/`, `node_modules/`, the local `.worktrees/`) without an explicit deny list at all. That would let you delete both `IGNORED_DIRECTORY_NAMES` and `IGNORED_RELATIVE_DIRECTORIES`, and the script would stay consistent with the SPDX runner. The trade-off is that untracked `.yml` files won't be linted until staged — but pre-commit already only protects staged paths, and "lint what's actually going into the repo" feels like the right semantics for a repo-scoped linter. If there's a deliberate reason to keep walking the working tree (e.g. so an unstaged-but-broken YAML still trips local `npm run lint:yaml`), worth a one-line comment in the script saying so.

2. **Asymmetry across linters.** The same `find … -not -path './.worktrees/*'` pattern lives in `lint:json` and the markdownlint invocation in `package.json`. Both also descend into `.claude/worktrees` for the same reason. This PR fixes only YAML. Fine to scope it that way, but worth either (a) acknowledging in the body that JSON/Markdown have the same characteristic or (b) opening a follow-up issue. The `git ls-files` approach in (1) would naturally fix all three at once.

3. **Asymmetric ignore list inside the script.** `IGNORED_DIRECTORY_NAMES` covers `.git`, `.worktrees`, `node_modules`, and `IGNORED_RELATIVE_DIRECTORIES` adds `.claude/worktrees`, but not `.claude` more broadly or `.codex/` (both gitignored at top level). Today neither contains `.yml`, so this is theoretical, but it's the kind of drift the explicit deny-list approach invites. Another argument for (1).

4. **Minor:** the `#!/usr/bin/env node` shebang on `scripts/lint-yaml.js` is dead weight since `package.json` invokes it as `node scripts/lint-yaml.js` and the file isn't chmod +x. Harmless; matches `check-spdx.js`'s pattern, so probably fine to leave.

5. **Minor:** `yaml.loadAll(content, () => {}, {})` passes an empty options object — equivalent to omitting the arg. Could drop the `{}` but it's not wrong.

Nothing here I'd block on. (1) is the only one I'd actually push back on before merge; everything else is fine as a follow-up or as-is.

JamesCropcho:
I took the `git ls-files` suggestion.

`lint:yaml` now runs `git ls-files -z -- '*.yml' '*.yaml' | node scripts/lint-yaml.js`, so the YAML linter works from the repo’s tracked file set instead of walking the working tree with an explicit deny list. That let me remove the ignore-list machinery from `scripts/lint-yaml.js` while keeping the same `js-yaml` parse diagnostics and filename-prefixed failure output.

I also updated `CONTRIBUTING.md` to say `lint:yaml` operates on git-tracked YAML files, refreshed the PR description to call out the YAML-only scope, and dropped the redundant empty `loadAll` options object while touching the script.

I left the JSON/Markdown follow-up out of this PR for now and just noted that scope in the PR body rather than broadening the change set here.

PR #356, opened 2026-04-27
Ignore nested worktree directories in repo-root scans

JamesCropcho:
## Summary
- exclude nested repo-local worktree directories from the repo-root JSON, Markdown, and ESLint scans
- ignore the top-level `./.worktrees/` and `./worktrees/` parking directories in Git
- preserve `main`'s newer tracked-file YAML lint during the rebase
- document that these exclusions apply to parent-checkout scans, not to checks run from inside a worktree

## Problem
The repo already had a partial defense against recursive tooling walking into nested sibling checkouts parked under the repository root, but it was incomplete:

- ESLint already ignored `.claude/` and `.worktrees/`, which incidentally covered `./.claude/worktrees/`, but it did not ignore a plain top-level `./worktrees/`.
- `lint:json` and `lint:markdown` still scanned the repo root broadly enough to pick up files from nested worktrees.
- The top-level `./.worktrees/` and `./worktrees/` directories were not ignored by Git, so a parent checkout could still accidentally stage those local parking areas even after lint stopped walking them.
- While rebasing, `main` had already moved `lint:yaml` to a git-tracked-file helper, so the conflict resolution needed to preserve that newer base-branch behavior rather than regress it.

That combination creates the kind of accidental cross-checkout coupling the task was worried about: a repo-root lint run can complain about files that belong to some other worktree, even though commits made inside that other worktree should be validated there, not from the parent checkout.

### Tooling
- `npm run lint:json`
  - switched to a `find ... -prune` form
  - prunes `./.git`, any `node_modules`, `./.claude/worktrees`, `./.worktrees`, and `./worktrees`
- `npm run lint:markdown`
  - now ignores `node_modules/**`, `.claude/worktrees/**`, `.worktrees/**`, and `worktrees/**`
  - I used `/**` forms intentionally after validating that directory-only ignores were not sufficient here
- ESLint flat config
  - kept the existing broad `.claude/` ignore
  - added the missing `worktrees/` ignore explicitly
- `.gitignore`
  - now ignores `/.worktrees/` and `/worktrees/` so local worktree parking directories cannot be accidentally staged from the parent checkout
- `npm run lint:yaml`
  - kept `main`'s newer `git ls-files -z -- '*.yml' '*.yaml' | node scripts/lint-yaml.js` form during conflict resolution
  - that already scopes YAML linting to the current checkout's tracked files, so it naturally avoids nested worktree bleed-through

### Docs
- updated `CONTRIBUTING.md` to say that repo-root recursive scans intentionally skip `./.claude/worktrees/`, `./.worktrees/`, and `./worktrees/`
- documented that `./.worktrees/` and `./worktrees/` are also ignored by Git from the parent checkout
- the doc note also makes the important non-goal explicit: when a command is run from inside a worktree, that worktree still validates itself normally

## Strategy And Rationale
I used a real Git worktree for the implementation instead of editing from the primary checkout because the behavior here is specifically about nested-checkout boundaries. That made it much easier to reason about whether the changes were preserving the right semantics.

A few choices were deliberate:

1. I did **not** change the Husky pre-commit script itself.
   - The hook dispatch logic only looks at the current checkout's staged files.
   - Broadening that logic would have increased risk without solving the actual recursive-scan problem.

2. I did **not** change `lint:spdx`, `typecheck`, `lint:shell`, `lint:dockerfile`, or `verify:build`.
   - `lint:spdx` operates on `git ls-files`, so it is already scoped to the current checkout's tracked files.
   - `typecheck`, `lint:shell`, `lint:dockerfile`, and `verify:build` already target explicit files rather than recursively walking the tree.
   - Changing checkout-local or explicit-target checks would have been broader than the task required.

3. I kept the existing `.worktrees/` exclusion instead of replacing it.
   - The request asked for `./.claude/worktrees/` and `./worktrees/`, but dropping the existing `.worktrees/` protection would have been a regression.

4. I added Git ignore entries only for the two local worktree parking roots.
   - `/.worktrees/` and `/worktrees/` are now protected from accidental parent-checkout staging.
   - I left `/.claude/` as-is because it was already ignored.

5. I switched JSON discovery to `-prune` instead of just adding more `-not -path` clauses.
   - That avoids descending into excluded trees at all.
   - It is a better fit for large nested worktree directories, where there can be thousands of files we never want to inspect from the parent checkout.

6. During the rebase, I kept `main`'s tracked-file YAML lint rather than carrying forward the older inline YAML scan from this branch.
   - That preserved the newer base-branch work.
   - It also aligns with the same core principle here: checks should operate on the current checkout's files, not wander into sibling worktrees.

7. I checked the ESLint anchoring concern empirically.
   - In the installed ESLint version, `.worktrees/` and `worktrees/` ignore the root directories, while `vendor/worktrees/` is not ignored.
   - Leading-slash forms like `/.worktrees/` and `/worktrees/` did not match the top-level paths, so I kept the working flat-config pattern.

## Why This Should Not Skip The Wrong Checks
This change is path-based for the recursive scanners that still walk the filesystem, tracked-file-based for YAML, and Git-ignore-based only for known local parking directories.

That distinction matters:
- Good behavior: "When a repo-root recursive scan sees these known nested worktree directories, skip them."
- Also good behavior: "When a check can work from `git ls-files`, scope it to the current checkout's tracked files."
- Also good behavior: "Keep parent-checkout `git add` from sweeping in local worktree parking directories."
- Bad behavior: "If this checkout is a worktree, skip checks."

The implementation does the first three, not the fourth.

So if someone commits **inside** a real Git worktree, that worktree still runs its own hook and its own lint commands against itself. The parent checkout just stops recursively wandering into sibling worktree directories that happen to live underneath it.

## Validation
I validated from a real branch-backed Git worktree checkout.

Checks run successfully after the feedback changes:
- `npm run lint:pre-commit`
- `git check-ignore -v .worktrees/example worktrees/example node_modules`
- ESLint `isPathIgnored()` probe for `.worktrees/example/file.js`, `worktrees/example/file.js`, `vendor/worktrees/example/file.js`, and `.claude/worktrees/example/file.js`

The full pre-commit run covers:
- `npm run verify:build`
- `npm run lint`
- `npm run lint:json`
- `npm run lint:markdown`
- `npm run lint:yaml`
- `npm run lint:dockerfile`
- `npm run lint:shell`
- `npm run lint:spdx`
- `npm run typecheck`


JamesCropcho:
## Review

Overall: sane, well-scoped, and the framing in "Why This Should Not Skip The Wrong Checks" is exactly right — this is a path-based exclusion of nested checkouts from repo-root recursive scans, not a worktree-state-based bypass. A commit made *inside* a worktree still runs that worktree's hook against its own files, which is what we want.

I poked at the changes empirically and read around the rest of the tooling. Findings below.

### What checks out

- The `find ... -prune -o` rewrite is correct and is also a nice efficiency improvement over the old `-not -path` chain — pruned subtrees are not descended into at all. I reproduced the behavior on a synthetic tree with `node_modules/`, `.git/`, `.worktrees/wt1/`, `.claude/worktrees/c1/`, and `worktrees/feature/inner/` populated with `*.json` files; only the top-level JSON files were emitted, all four pruned subtrees were silent.
- `markdownlint --ignore 'worktrees/**'` (and the other three globs) does what the PR claims; the previous directory-only form was indeed weaker. Matches the PR's stated reason for the glob switch.
- Switching the `find` predicate from `-not -path '*/node_modules/*'` to `-name node_modules -prune` is a strict improvement: it still catches nested `node_modules/` at any depth and stops descent rather than filtering after the fact.
- The "did not change" list is well-justified: `lint:spdx` is scoped via `git ls-files`, `typecheck` uses an explicit `include` list in `.tsconfig.checkjs.json`, and `lint:dockerfile` / `lint:shell` / `verify:build` all target explicit files. I checked these and didn't find any other recursive scanners that would silently pick up nested checkouts.
- Not touching the Husky dispatcher is the right call. It only inspects `git diff --cached`, which is naturally scoped to the current checkout.

### Worth flagging (not blockers)

1. **`.gitignore` does not exclude `/worktrees/` or `/.worktrees/`.** Your point 4 in the rationale explains the choice, and I agree it's a separate policy decision. Worth being explicit that the consequence is: a `./worktrees/feature/` checkout will not show up as lint noise from the parent, but it *can* still be accidentally `git add`ed from the parent. If that ever becomes a real footgun, a follow-up PR adding the `.gitignore` entries (with the same scoping argument) would be the natural place — keeping it out of this PR is fine.

2. **ESLint ignore patterns are unanchored, the `find` paths are anchored.** In the flat config `{ ignores: ['.claude/', '.worktrees/', 'worktrees/'] }`, those patterns match at any depth (e.g., a hypothetical `vendor/foo/worktrees/`), whereas the `find -path './worktrees'` and the markdownlint `worktrees/**` are anchored at the repo root. Almost certainly fine in practice and arguably more defensive for ESLint, but the inconsistency is mildly surprising — if you want symmetry, `'/worktrees/'` and `'/.worktrees/'` would mirror the others. Either way, no functional issue today.

3. **`-path './.git'`** matches only the top-level `.git` (same as the old `-not -path './.git/*'`). A nested non-worktree `.git` somewhere in the tree wouldn't be pruned, but that's a pre-existing property and nothing in this PR makes it worse — and the realistic case (nested `.git` inside a worktree) is already covered because the parent worktree directory is pruned.

### Omissions

None that I can see. The set `{ .claude/worktrees, .worktrees, worktrees }` covers the conventional locations, and every recursive scanner in the repo is touched. Non-recursive checks (spdx, typecheck, docker, shell, build) genuinely don't need the change.

### Should anything be split out

- The `find` refactor (from `-not -path` to `-prune`) is technically tangential to "ignore two more paths," but it's directly motivated by the markdownlint discovery and the prune form is what makes adding multiple exclusions clean. Bundling is the right call.
- The `CONTRIBUTING.md` paragraph belongs here.
- The `.gitignore` question is the one thing I'd happily see deferred to a separate PR if you decide to do it at all.

LGTM in spirit; the flagged items are observations, not requested changes.

JamesCropcho:
Thanks for the careful pass here. I made the feedback-driven change I agree with after the follow-up discussion: `/.worktrees/` and `/worktrees/` are now in `.gitignore`, and `CONTRIBUTING.md` now calls out that those two parent-checkout parking directories are ignored by Git as well as skipped by repo-root recursive scans.

I also checked the ESLint anchoring observation empirically with the installed ESLint API. In this flat-config setup, the existing `.worktrees/` and `worktrees/` ignore entries do ignore the top-level directories, while `vendor/worktrees/` is not ignored. The leading-slash forms (`/.worktrees/`, `/worktrees/`) did not match the top-level paths in this ESLint version, so I left the working ESLint pattern in place rather than changing it for symmetry.

Validation after the change:

- `npm run lint:pre-commit`
- `git check-ignore -v .worktrees/example worktrees/example node_modules`
- ESLint `isPathIgnored()` probe for `.worktrees/example/file.js`, `worktrees/example/file.js`, `vendor/worktrees/example/file.js`, and `.claude/worktrees/example/file.js`

JamesCropcho:
Re-reviewed the current diff and ran the same kind of probe locally to settle the ESLint question for myself.

Both addressed items land cleanly:

- `.gitignore` now has `/.worktrees/` and `/worktrees/` with the inline comment tying them to the lint exclusions, and the `CONTRIBUTING.md` paragraph closes the loop by saying both things in one place. That's the version we're keeping.
- I reproduced the ESLint anchoring probe against the installed ESLint with this exact `{ ignores: ['.claude/', '.worktrees/', 'worktrees/'] }` config. Confirmed: the unanchored form correctly ignores the top-level directories and does **not** match `vendor/worktrees/x/c.js`, while the leading-slash forms `/.worktrees/`, `/worktrees/`, `/.claude/` match nothing in this version. So my earlier "unanchored, may match `vendor/foo/worktrees/`" worry was wrong in practice for flat-config ESLint — the working pattern is the right one to keep, and switching to leading-slash for "symmetry" would have actively broken the ignore. Good empirical catch.

One side observation while re-reading: the rebased `lint:yaml` (now `git ls-files -z -- '*.yml' '*.yaml' | node scripts/lint-yaml.js`) is a strictly stronger form of the same scoping principle — tracked-file-driven rather than `find`-driven — and removes any need for worktree prune entries on the YAML side. Nice that it lined up that way.

Nothing else outstanding from my end.

PR #357, opened 2026-04-27
Add internal MongoDB Node driver adapter

JamesCropcho:
## Summary
- add a private `mongo-driver/adapter.js` boundary that analyzes a MongoDB Node driver `Collection` or `Db` without shelling out
- keep the first adapter scope analysis-only by rejecting formatter, persistence, plugin, and shell-only options clearly
- add focused `mongo-driver` tests plus contributor docs for the new internal boundary

## Why
- continue the #263 groundwork after #332, #339, and #346 without committing to a public package/export shape yet
- prove the Node-driver path can reuse the shared config and engine layers while preserving driver-native BSON behavior
- keep this PR narrowly useful even if the eventual public API shape changes later

## Validation
- `npm run lint -- --quiet`
- `npm run typecheck`
- `npm run lint:markdown`
- `npm run lint:spdx`
- pre-commit hook: `npm run lint:pre-commit`
- focused container-backed test using the existing local Variety test image:
  - `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/mongo-driver/AdapterTest.js`

Refs #263 (@JamesCropcho).

### DIFF
--- a/.eslint.config.js
+++ b/.eslint.config.js
@@ -84,7 +84,7 @@ module.exports = (async () => {
       },
     },
     {
-      files: ['.eslint.config.js', 'bin/variety', 'build.js', 'cli/**/*.js', 'mongo-shell/transport-options.js', 'mongo-shell/launcher.js', 'test/**/*.js'],
+      files: ['.eslint.config.js', 'bin/variety', 'build.js', 'cli/**/*.js', 'mongo-driver/**/*.js', 'mongo-shell/transport-options.js', 'mongo-shell/launcher.js', 'test/**/*.js'],
       ignores: ['test/fixtures/**/*.js'],
       rules: nodeModernizationRules,
     },
--- a/.tsconfig.checkjs.json
+++ b/.tsconfig.checkjs.json
@@ -25,6 +25,7 @@
     "cli/**/*.js",
     "core/option-validation.js",
     "core/config.js",
+    "mongo-driver/**/*.js",
     "mongo-shell/transport-options.js",
     "mongo-shell/launcher.js",
     ".eslint.config.js",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,8 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
 
 ## Repo Layout and the `variety.js` Build
 
-`variety.js` at the repo root is a generated file, assembled from seven sources in this order:
+The repo is organized around these sources and layers. The generated `variety.js`
+bundle is assembled from the shell-compatible sources among them, in build order:
 
 - `core/formatters/ascii.js` — built-in ASCII table formatter. Self-contained
   IIFE; registers an `ascii` factory on `shellContext.__varietyFormatters`.
@@ -68,6 +69,13 @@ As an additional (not required) dependency, [Docker](https://www.docker.com/) or
   The only place that touches Node `spawnSync` for spawning `mongosh`/`mongo`.
   Future interfaces (MCP, etc.) can import from `mongo-shell/` without
   reaching into `cli/`.
+- `mongo-driver/adapter.js` — internal MongoDB Node driver adapter that runs
+  Variety analysis directly against a `Collection` or `Db` without shelling out.
+  Reuses `core/config.js` and `core/engine.js`, preserves driver-native BSON
+  values, and rejects formatter, persistence, plugin, and shell-only options so
+  this boundary stays analysis-only for now. Node-only; not compiled into
+  `variety.js`. A future `@variety/node` or `@variety/mcp-server` package could
+  grow on top of this layer.
 - `cli/main.js`, `cli/options.js` — Node-side CLI parsing and compatibility
   handling. Depends on `core` and `mongo-shell/`; this is the future
   `@variety/cli` package boundary.
@@ -79,7 +87,8 @@ has no runtime deps on any other layer. `core/analyzer.js` depends on
 `core/engine.js` plus `core/formatters/`. `mongo-shell/adapter.js` depends on
 `core/config.js` and `core/analyzer.js`. `mongo-shell/transport-options.js`
 depends on `core/option-validation.js`; `mongo-shell/launcher.js` has no
-`core` dep. `cli/options.js` depends on `core/config.js` and
+`core` dep. `mongo-driver/adapter.js` depends on `core/config.js` and
+`core/engine.js`. `cli/options.js` depends on `core/config.js` and
 `mongo-shell/transport-options.js`; `cli/main.js` depends on `cli/options.js`
 and `mongo-shell/launcher.js`. `build.js` composes `core/formatters/` +
 `core/option-validation.js` + `core/config.js` + `core/engine.js` +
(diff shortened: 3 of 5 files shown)


JamesCropcho:
Read through the diff and the surrounding `core/config.js` / `core/engine.js` / `mongo-shell/adapter.js` boundary. Overall this looks sane and well-scoped — a sensible next step on the #263 thread that doesn't paint anything into a corner.

## Gestalt

- Scoping is right. Keeping this private (no entry in `package.json`'s `files`, no `exports` map, no README example yet) and explicitly rejecting formatter/persistence/plugin/shell-only options is exactly the kind of "first adapter scope" decision that lets the public shape stay open. The `UNSUPPORTED_OPTION_REASONS` map with a per-option *why* is a really nice UX touch.
- The layering matches the existing pattern: `mongo-driver/adapter.js` reuses `core/config.js` + `core/engine.js` and stays parallel to `mongo-shell/adapter.js` rather than building a new abstraction. CONTRIBUTING.md updates accurately reflect that.
- Tests cover the load-bearing cases (real Mongo connection, `Db` + name overload, driver-native BSON wrapper preservation via `BSONSymbol`, rejection paths, unknown-key path). Good shape.

## Smaller things worth considering — none of which I'd block on

1. **The post-iteration `countMatchingDocuments` looks redundant.** In `analyzeCollection` you compute `documentsCount` after the loop with `countMatchingDocuments(typedCollection, config.query, config.limit)`. But the cursor already yields exactly `min(matches, limit)` documents (or all matches when `limit === 0`), so the iteration count and that second count are equal by construction. You could track `documentsCount` inside the `for await` loop and drop the second roundtrip — small perf win, and it eliminates the (admittedly tiny) race window where docs are inserted/deleted between iteration and recount, skewing `percentContaining`. `core/engine.js` already has an `analyzeDocuments(config, documents, { documentsCount, log })` helper that does roughly this, but it uses a sync `for...of`, so you can't pass an async cursor — the manual loop is unavoidable, just not the second count.

2. **Two error messages for the same shape check.** `ensureOptionsObject` throws `'MongoDB Node driver analysis options must be an object.'` and then `validateAnalysisOptions` runs its own `ensureInputObject` which throws `'Analysis options must be an object.'`. The first one wins, but the duplication is a small drag. `assertSupportedOptions` is the part that genuinely has to run before validation; the object-shape check could be left to `validateAnalysisOptions`.

3. **`readQuery` is duplicated from `core/config.js`.** Trivial, but if `core/config.js` ever exported it, you could drop the local copy.

4. **No central registry tying `SUPPORTED_ANALYSIS_OPTION_NAMES` to `core`.** If someone adds a new option to `ANALYSIS_OPTION_DESCRIPTORS` in [core/config.js](core/config.js), this adapter will silently report it as `Unknown MongoDB Node driver analysis option`. That's the *correct* failure mode for now (analysis-only intent), but a brief comment in [mongo-driver/adapter.js](mongo-driver/adapter.js) pointing readers at `core/config.js` when they extend either side would save a future-you a minute.

5. **Empty-cursor path isn't directly exercised.** The fake-collection rejection tests use `[]` and bail before iteration, but there's no positive test that `analyzeCollection` against zero matching docs returns `[]` cleanly. Probably fine — the engine handles it — but it's the boundary where `documentsCount === 0` and the percentages divide.

6. **`config.sort` cast to `import('mongodb').Sort`.** Necessary because `core/config.js` types it as `Record<string, unknown>` while the driver type is more constrained. Worth a one-line type-narrowing helper later if other places need to make the same cast.

## Things I think are correctly *out* of this PR

- Streaming results / progress callbacks (mentioned as an open question in #263) — fits a later PR once the boundary settles.
- README example — issue #263's acceptance asks for one, but this PR is explicit about staying internal, so deferring is right.
- A `@variety/node` / `@variety/mcp-server` package shape — premature until a second consumer materializes.
- `secondaryOk` rejection: correct; read preference is properly the caller's `MongoClient` concern.

Refs #263.

JamesCropcho:
Thanks for the review. I took the changes that looked like they tightened the adapter without broadening the scope:

- Removed the post-iteration count and now track the analyzed document count during the async cursor pass.
- Dropped the adapter-local options object-shape guard so shared `validateAnalysisOptions` owns that validation.
- Added a short note to keep the driver adapter option allowlist aligned with `core/config.js` as shared analysis options grow.
- Added a positive empty-cursor test.

I intentionally left the `readQuery` duplication alone for now since exporting it would broaden the shared config API for a small internal helper, and I left the `config.sort` cast as-is until there is a second use that justifies a helper.

Validation run after rebasing onto the updated branch:

- `npm run lint -- --quiet`
- `npm run typecheck`
- focused container-backed `test/cases/mongo-driver/AdapterTest.js`

JamesCropcho:
Re-read the updated diff at 6115eb6. The four changes look right:

- **Documents-count tracked in the loop** ([mongo-driver/adapter.js](mongo-driver/adapter.js)): clean. One round-trip removed, race window closed, and `analyzeCollection` reads more linearly now.
- **Shape guard dropped**: good. The reordering — `validateAnalysisOptions(options)` first, then `assertSupportedOptions(source)` against the raw `options` — is the right shape, since [core/option-validation.js:146-179](core/option-validation.js#L146-L179) only copies *known* descriptor keys onto its result, so the unknown-key detection has to run against the raw input. Confirms why `validatedOptions` couldn't be reused there.
- **Allowlist alignment comment**: short and load-bearing. Future-me thanks you.
- **Empty-cursor test**: covers the `documentsCount === 0` boundary and is now the simplest call site for the fake collection — nice.

Your calls to leave `readQuery` and the `Sort` cast alone both make sense; both would be speculative API surface for a single internal use.

One very small observation, take or leave: with the new order, an unsupported option passed with a *malformed* value (e.g. `outputFormat: 123`) will surface the core type error (`outputFormat must be a string`) rather than the friendlier `does not support "outputFormat"`. The existing rejection tests don't hit this since they use valid types. Not worth churning for — the user is wrong either way and the type error still tells them they're wrong — but if you ever want the "does not support" message to win unconditionally, the fix is to gate `assertSupportedOptions` behind a small `isPlainObject(options)` check and run it before validation.

Otherwise this looks ready from my side.

JamesCropcho:
I took the small unsupported-option precedence tweak from the latest feedback.

What changed:

- Added a plain-options-object guard in `mongo-driver/adapter.js` so unsupported/unknown option checks can run before shared value validation when the input is actually an options object.
- Kept malformed non-object inputs on the shared `Analysis options must be an object.` validation path.
- Added a regression test for `outputFormat: 123` to ensure the adapter still reports `does not support "outputFormat"` instead of the lower-level core type error.

No other changes from that comment seemed needed.

Validation:

- `npm run lint -- --quiet`
- `npm run typecheck`
- focused container-backed `test/cases/mongo-driver/AdapterTest.js`
- pre-commit `npm run lint:pre-commit`

JamesCropcho:
Verified at 2486257. The `isPlainOptionsObject` gate plus running `assertSupportedOptions` before `validateAnalysisOptions` gives the right precedence: unsupported keys win regardless of the value's type, while non-object inputs still fall through to the shared `Analysis options must be an object.` path. Walked the cases (`undefined`, plain object with bad-type unsupported key, plain object with unknown key, `null`, array, string) and they all land on the message you'd want.

The new `outputFormat: 123` regression test pins the specific behavior I was worried about, so this won't quietly drift later.

One purely stylistic micro-nit, take or leave: the gate could just be `if (isPlainOptionsObject(options)) { assertSupportedOptions(options); }` — when `options` is `undefined`, `assertSupportedOptions({})` is a no-op anyway, so the `?? {}` and the explicit `undefined` branch are doing extra work for the same outcome. Not worth a commit on its own.

This looks ready from my side. Nice iteration loop.

PR #365, opened 2026-05-07
Add ESLint result cache to speed up warm lint runs

JamesCropcho:
## Summary

- `lint:eslint` now passes `--cache --cache-strategy content`
- `.eslintcache` added to `.gitignore`

## Motivation

Measured on this repo (50 JS files, 9 MD files):

| Run | Time |
|---|---|
| Cold (no cache) | ~5 s |
| Warm (cache, no files changed) | ~1.2 s |

The speedup comes mainly from skipping the TypeScript language service startup that `recommendedTypeChecked` requires on the test files — on a warm run, ESLint sees unchanged content hashes and skips those files entirely.

`--cache-strategy content` hashes file contents rather than relying on mtime/size, so the cache stays valid across branch switches and worktree copies without stale hits.

The cache file lives at `.eslintcache` in the repo root and is safe to delete (ESLint rebuilds it on the next run). It is gitignored.

## Test plan

- [x] Pre-commit hook passed (all checks green, cache file created)
- [ ] Run `npm run lint:eslint` twice; second run should be noticeably faster


JamesCropcho:
Overall: this looks sane to me, and I think it is a reasonable PR. It is small, local-only, and `--cache-strategy content` is the right choice for this repo because branch/worktree churn is exactly where mtime-based caching gets annoying. I do not see this as something that needs to be split out; the script change and `.gitignore` entry belong together.

The one thing I would tighten before merging is the workflow alignment with `lint:eslint:fix`. Right now `lint:eslint` would create `.eslintcache`, but `lint:eslint:fix` still invokes ESLint without `--cache`; ESLint deletes the default cache file when it is run that way. That is not dangerous, but it means a fix run quietly blows away the warm-cache benefit. I would either mirror the same cache flags onto `lint:eslint:fix` as well, or call that out as intentional. ESLint already reprocesses cached files with messages when `--fix` is enabled, so using the cache for the fix script should be safe.

I would also add a very small note in `CONTRIBUTING.md` under the ESLint/linting section: `npm run lint:eslint` uses the repo-root `.eslintcache`, it is ignored, and it is safe to delete if lint behavior ever seems stale. I do not think README needs anything here; this is contributor tooling, not user-facing behavior.

Tiny nit: the `.gitignore` comment may be a little over-specific with "content-hashed" baked into it, since that could go stale if the script changes later. Not a blocker, just the kind of detail I usually prefer to keep in the script/docs rather than the ignore file.

So: good idea, low risk, worth doing. I would just make the cache behavior consistent across the normal ESLint scripts and give contributors the one-sentence escape hatch.

JamesCropcho:
All three suggestions applied in the follow-up commit:

- **`lint:eslint:fix`** now carries the same `--cache --cache-strategy content` flags as `lint:eslint`, so a fix run updates the cache rather than bypassing it.
- **CONTRIBUTING.md** has a one-sentence note under the Shared Baseline section: delete `.eslintcache` if lint behavior ever seems stale.
- **`.gitignore` comment** simplified — "content-hashed" removed so it won't go stale if the strategy ever changes.

PR #366, opened 2026-05-07
Use service containers for modern MongoDB CI tests

JamesCropcho:
## Summary
- run the MongoDB 8.2/8.0/7.0 matrix lanes and Node 24 smoke test against GitHub Actions MongoDB services instead of building custom test images
- add a mounted Docker mongosh client wrapper plus MONGO_SHELL_COMMAND override for shell-backed tests
- keep MongoDB 5.0 on the legacy container path; that lane now reuses mounted node_modules when available and falls back to an in-container npm install with HUSKY disabled when not
- document the service-client image reuse assumption and update CONTRIBUTING.md for the new CI layout

## Validation
- npm run lint:yaml
- npm run lint:shell
- npm run lint:markdown
- npm run lint:eslint
- npm run lint:json
- npm run lint:spdx
- npm run typecheck
- git diff --check
- MONGO_SHELL_COMMAND=/bin/echo smoke check for the shell override
- MONGODB_VERSION=5.0 npm run test:container (136 passing)
- git commit hook ran npm run lint:pre-commit successfully

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,10 +82,48 @@ jobs:
     needs: changes
     if: needs.changes.outputs.only-markdown != 'true'
     runs-on: ubuntu-latest
+    services:
+      mongodb:
+        image: mongo:${{ matrix.mongodb-version }}
+        ports:
+          - 27017:27017
+        options: >-
+          --health-cmd "mongosh --quiet --eval 'db.adminCommand({ ping: 1 }).ok'"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 20
     strategy:
       fail-fast: false
       matrix:
-        mongodb-version: ['8.2', '8.0', '7.0', '5.0']
+        mongodb-version: ['8.2', '8.0', '7.0']
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: '22'
+          cache: npm
+      - run: npm ci
+        env:
+          HUSKY: '0'
+      - name: Start Mongo shell client
+        run: |
+          # The mongodb service has already pulled this image; avoid a duplicate registry pull.
+          docker run --detach --rm --pull never \
+            --name variety-mongosh-client \
+            --network host \
+            --volume "$PWD:$PWD" \
+            --workdir "$PWD" \
+            mongo:${{ matrix.mongodb-version }} \
+            tail -f /dev/null
+      - run: npm run test:mocha
+        env:
+          MONGO_SHELL_COMMAND: ./test/bin/mongosh-in-service-container.sh
+
+  test-legacy:
+    name: Test (Node 22, MongoDB 5.0)
+    needs: changes
+    if: needs.changes.outputs.only-markdown != 'true'
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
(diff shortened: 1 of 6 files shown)


JamesCropcho:
Read through the diff and surrounding code (`MongoShellRunner.js`, `VarietyHarness.js`, `test-in-container.sh`, the existing CI). Overall: this is a sane, well-scoped refactor and I'd merge it. A few observations, mostly nits.

## Gestalt: yes, this is the right move
Service containers + native runner Node is the standard GHA pattern for DB tests. Building a custom Dockerfile per matrix lane just to get `mongosh` and Node next to the DB was overkill, and `actions/setup-node` already gives you a faster, cached Node toolchain than NVM-inside-a-container ever will. Keeping 5.0 on the old path because it's the only lane that needs the legacy `mongo` binary is exactly the right seam to cut on.

The `MONGO_SHELL_COMMAND` escape hatch is a clean abstraction — it's a one-line addition to `MongoShellRunner.js` with no behavior change for the default path, and it's reusable outside CI (e.g. local development against a remote MongoDB).

## Things I'd flag

**Sidecar `--pull never` + `--network host` is correct but load-bearing.** It works because (a) the GHA service-container declaration pulls `mongo:<version>` first, so the image is in the local daemon by the time the `Start Mongo shell client` step runs, and (b) hosted Linux runners support host networking. Both are durable assumptions for `ubuntu-latest`, but neither is documented. A one-line comment in the workflow explaining *why* `--pull never` is safe would save a future reader 10 minutes.

**The legacy (5.0) lane silently changed its install behavior.** Before this PR, `docker/init.sh` always ran `npm install` *inside* the container. Now, because the runner does `npm ci` before invoking `test:container` and the worktree is bind-mounted, the `[ -x node_modules/.bin/mocha ]` branch fires and the container reuses the runner-built `node_modules`. That's almost certainly fine (both runner and container are Linux x64 glibc Ubuntu, and variety's runtime deps are pure JS), but it is a behavior change for the 5.0 lane that isn't called out in the PR description. If any transitive dep ever ships a native binding compiled against a different ABI than the container's, this lane will surface it first. Worth a sentence in the PR body.

**`export HUSKY=0` in `init.sh` is belt-and-suspenders.** With the new conditional, `npm install` only runs in the container when `node_modules` isn't mounted — and that path was already the same as before this PR (where it presumably worked without `HUSKY=0`). Not wrong, just slightly redundant; if you want to keep it, a half-line comment explaining the case it covers would help.

**`MONGO_SHELL_COMMAND` early-return bypasses the cached promise.** Trivial, but it means each call re-reads `process.env`. Negligible cost; mentioning only because the symmetry of the surrounding code suggests it was unintentional.

**The wrapper hardcodes `mongosh`.** If you ever want to exercise the *legacy* `mongo` shell against a service container (e.g. some future MongoDB version where both ship), the wrapper can't do it as-is. Today this is moot — only 5.0 needs `mongo` and 5.0 stays on the Dockerfile path — but a `MONGO_SHELL_BINARY` env (defaulting to `mongosh`) would future-proof it for ~zero cost. Probably YAGNI; flagging for the record.

## Possibly out of scope (but small enough to keep)
- `HUSKY=0` + the `node_modules` reuse in `init.sh` are independent improvements to the legacy path that ride along with the service-container work. Bundled is fine — they're tiny and the story holds together — but they're the only changes here that aren't strictly about the matrix split.

## Omissions worth a moment
- `CONTRIBUTING.md` documents the *CI* shape but doesn't give a contributor a recipe for reproducing a service-container lane locally (`docker run mongo:8.2`, set `MONGO_SHELL_COMMAND`, run `npm run test:mocha`). Up to you whether that's worth adding — `npm run test:container` is still the documented local path.
- No cleanup/`docker stop` of `variety-mongosh-client`. Fine for ephemeral hosted runners; would matter on self-hosted. Not worth fixing speculatively.

## Things that are *not* problems (verified)
- Wrapper script's `$WORKDIR` / `$CONTAINER` interpolations are safe — both go to `docker exec` flag args, no shell re-parsing.
- The `lint:shell` script picks up the new wrapper.
- SPDX headers are correct on the new file.
- `required-ci-gate` was updated for the new `test-legacy` job in all three places (needs, env, and the markdown-only and full-CI assertion blocks). Easy to miss one of those; you didn't.
- README has no references to the old matrix shape, so no README sync is needed.

Net: ship it. The flags above are all "would be nicer" rather than "should block."

JamesCropcho:
Addressed the feedback items I agreed should change here:

- Added workflow comments explaining why `--pull never` is safe: the MongoDB service has already pulled the same `mongo:<version>` image.
- Added a comment in `docker/init.sh` explaining the `HUSKY=0` fallback-install case.
- Moved `MONGO_SHELL_COMMAND` through the same cached promise path as normal shell discovery.
- Updated the PR body to explicitly call out that the MongoDB 5.0 lane now reuses mounted `node_modules` when available and falls back to in-container install otherwise.

I intentionally left `MONGO_SHELL_BINARY`, a local service-container reproduction recipe, and explicit client-container cleanup out of this PR. Those still read like future-proofing/self-hosted polish rather than necessary changes for this hosted-runner CI split.

Validation after the follow-up edits:
- `npm run lint:yaml`
- `npm run lint:shell`
- `npm run lint:eslint`
- `npm run typecheck`
- `git diff --check`
- `MONGO_SHELL_COMMAND=/bin/echo` smoke check for the shell override
- amend commit hook reran `npm run lint:pre-commit` successfully

JamesCropcho:
Re-read the four follow-up changes — all look good:

- The `--pull never` comment is short and says exactly the right thing (the *why*, not the *what*). Both `test` and `test-smoke` got it.
- The `HUSKY=0` comment frames it as "fallback-install hook avoidance" rather than describing the line below it, which is what makes it useful.
- The `MONGO_SHELL_COMMAND` rework now goes through the same `mongoShellCommandPromise` cache via `Promise.resolve(...)`. Cleaner than my original suggestion of "just live with the asymmetry" — both branches now share one memoized resolution.
- PR body now flags the 5.0-lane `node_modules` reuse explicitly. Good.

Agreed on all three deferrals: `MONGO_SHELL_BINARY`, local-repro recipe, and explicit client-container cleanup are speculative until something actually needs them. No reason to pre-pay.

Ship it.

PR #367, opened 2026-05-08
Add shell and driver parity tests

JamesCropcho:
## Summary

- add a dedicated `test/cases/parity/` suite that compares `variety.js` shell JSON output with `mongo-driver/adapter.js` structured results
- add a small parity-specific fixture that exercises shared analysis semantics across nested objects, arrays, filtering, sorting, limits, examples, and last values
- document the new parity test group in `CONTRIBUTING.md`

## Context

This is the parity-test follow-up discussed as the next low-waste step toward both #263 and #264.

It builds on the earlier groundwork:

- #332 extracted the reusable internal analysis engine.
- #339 centralized shared analysis config/default handling.
- #346 split shell adapter preparation from execution as preparation for #264.
- #357 added the internal MongoDB Node driver adapter for #263.

Together those PRs leave two important analysis entrypoints in place: the existing shell JSON path and the internal Node driver adapter. This PR checks that they return the same normalized structured rows for supported analysis options before either #263 or #264 grows more public API surface.

## Scope

The matrix intentionally focuses on analysis-only behavior:

- default options
- `query`, `sort`, and `limit`
- `maxDepth`
- `arrayEscape`, `compactArrayTypes`, `excludeSubkeys`, and `showArrayElements`
- `lastValue` and `maxExamples`

It deliberately avoids persistence, plugins, formatter behavior beyond requesting shell JSON, and shell-only transport options. That is intentional because #357 explicitly keeps the Node driver adapter analysis-only for now and rejects formatter, persistence, plugin, and shell-only options.

## Validation

- `npm run lint:pre-commit`
- focused MongoDB 8.2 / Node.js 22 container parity test:
  - `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/parity/ShellDriverParityTest.js`
- focused MongoDB 5.0 / Node.js 22 legacy-shell parity test:
  - `./node_modules/.bin/mocha --reporter spec --timeout 15000 test/cases/parity/ShellDriverParityTest.js`

Refs #263 and #264.


JamesCropcho:
Overall this looks sane and well-placed in the #332 → #339 → #346 → #357 sequence. The matrix matches the analysis-only allowlist that #357 deliberately enforces, the deep-equal-after-`JSON.parse(JSON.stringify(...))` normalization is the right way to bridge BSON wrappers vs. shell JSON output, and the fixture exercises the option semantics it advertises. CI is green across the modern lanes and the legacy MongoDB 5.0 / `mongo` shell lane, which is exactly the parity I'd want this test to cover.

A handful of things I'd consider before merging — none of them blockers:

**Worth catching while you're here**

- The CONTRIBUTING.md list you're updating still does not enumerate [test/cases/mongo-shell/](test/cases/mongo-shell/) (`AdapterValidationTest.js`, `SecondaryReadsTest.js`). That's a pre-existing oversight, but since this PR is editing exactly the sentence that enumerates `test/cases/...` directories, it's the natural place to fold the fix in rather than leaving it as a separate one-line PR later.

**Small simplifications, take them or leave them**

- [analyzeWithShell](test/cases/parity/ShellDriverParityTest.js#L34-L48) hand-rolls the JSON parse + array-shape check that [VarietyHarness.runJsonAnalysis](test/helpers/VarietyHarness.js#L92-L101) already does. `AnalysisResultsValidator.results` is a public field, so `(await test.runJsonAnalysis(options)).results` collapses that helper to one line. The current form is fine; it just duplicates harness logic.
- The \"array and excluded-subkey options\" case bundles four flags (`arrayEscape`, `compactArrayTypes`, `excludeSubkeys`, `showArrayElements`). Convenient now, slightly noisier diagnostics if one of them ever drifts. Consider splitting if you expect failures here to be the kind you want to bisect quickly.

**Coverage gap to think about (probably a follow-up, not this PR)**

- The fixture is type-homogeneous per field — every `rank` is a Number, every `name` is a String, every `tags` is an Array. The most likely place for shell↔driver parity to break is BSON type classification, so a fixture with at least one field that mixes types across documents (e.g. `score` being Number in some, String in others, plus a `Binary` somewhere) would harden this against the regression class it's most aimed at. Reasonable to defer; flagging it because if I were going to extend this matrix later, that's where I'd start.

**Scope**

I don't see anything in here that wants to be a separate PR. Keeping it scoped to analysis-only parity (no formatter / persistence / plugin entry points) is the right call given #357's stance, and the PR description correctly frames this as the prerequisite for #263 / #264 rather than as part of either.

**Nits**

- `normalizeRows` types its return as `unknown` via the JSDoc; `VarietyResultRow[]` would be more honest, though `deepEqual` doesn't care.
- SPDX headers are correct (single year 2026, both tags).

JamesCropcho:
Addressed the actionable pieces I agreed with in `8d00e38`:

- Added `test/cases/mongo-shell/` to the `CONTRIBUTING.md` test-directory list while touching that same sentence.
- Simplified `analyzeWithShell()` to use `VarietyHarness.runJsonAnalysis(...).results` instead of duplicating JSON parsing and array-shape validation.
- Split the combined array/exclude-subkey parity case into separate `array option interactions` and `excluded-subkey options` cases for clearer failure diagnostics.
- Tightened `normalizeRows()` so its JSDoc return type is `VarietyResultRow[]`.

I left the mixed-type/Binary fixture expansion out of this PR because the feedback framed that as probably follow-up territory, and I agree it is a better next hardening pass than something to fold into this already-scoped parity PR.

Validation after the changes:

- `npm run lint:pre-commit`
- focused MongoDB 8.2 / Node.js 22 parity test: `6 passing`
- focused MongoDB 5.0 / Node.js 22 legacy-shell parity test: `6 passing`

JamesCropcho:
Re-checked everything in `8d00e38`:

- [CONTRIBUTING.md](CONTRIBUTING.md#L114) now lists [test/cases/mongo-shell/](test/cases/mongo-shell/) in alphabetical order alongside the other directories — clean.
- [analyzeWithShell](test/cases/parity/ShellDriverParityTest.js#L34-L40) now delegates through `VarietyHarness.runJsonAnalysis(...).results`. The `results.results` chain reads slightly oddly because the local variable holds an `AnalysisResultsValidator`, but that's purely cosmetic and not worth churning over.
- The split into separate `array option interactions` and `excluded-subkey options` cases is what I had in mind — better isolation if either set ever drifts.
- [normalizeRows](test/cases/parity/ShellDriverParityTest.js#L51-L60) is now typed as `VarietyResultRow[]`; the internal `unknown` bridge matches the cast style elsewhere in the file. Good.

Agree with deferring the mixed-type / Binary fixture expansion — that's a hardening pass that deserves its own PR rather than scope creep here.

Modern + legacy MongoDB lanes are still green. From my side this is good to merge.

Issue #368, opened 2026-05-08
Investigate full-suite MongoDB timeouts in container Mocha runs


JamesCropcho:
## Problem

While validating #367 from a worktree, a full ad hoc Mocha-suite run inside the existing MongoDB 8.2 / Node.js 22 Variety test image reached `115 passing`, then cascaded into Mongo-related failures and timeouts across unrelated suites.

This did not reproduce in the focused parity tests added by #367:

- MongoDB 8.2 / Node.js 22 focused parity test passed.
- MongoDB 5.0 / Node.js 22 legacy-shell focused parity test passed.

That makes this look like a full-suite container/harness or Mongo lifecycle reliability issue rather than a parity-test assertion failure.

## Observed Symptoms

The full-suite run failed after many earlier tests had already passed. The first visible assertion failure was in `ParametersParsingTest.js`:

```text
AssertionError [ERR_ASSERTION]: expected error.stdout to include:
The collection specified (--unknown--) in the database specified (test) does not exist or is empty.
```

After that, unrelated hooks began timing out:

- `Parameters parsing` afterEach hook
- `Formatter registry dispatch` beforeEach hook
- `MongoDB Node driver adapter` beforeEach hook
- `Secondary reads` beforeEach hook
- `shell and MongoDB Node driver analysis parity` beforeEach hook
- `Continuous logging` beforeEach hook
- `Persistence of results` beforeEach hook
- `Plugins` beforeEach hook

The run then needed manual container cleanup; `podman stop` eventually escalated to `SIGKILL`.

## Reproduction Context

This was observed from PR #367 using an ad hoc container command equivalent to:

```bash
podman run --rm --tty \
  --volume /home/j/variety/.worktrees/shell-driver-parity:/opt/variety:z \
  --entrypoint bash \
  localhost/variety:1.5.2-mongodb8.2-nodejs22 \
  /tmp/run-variety-focused-test.sh 'test/**/*Test.js'
```

The helper script mirrored the normal container setup pattern:

- start `mongod`
- wait for `db.adminCommand({ ping: 1 })`
- run `./node_modules/.bin/mocha --reporter spec --timeout 15000 'test/**/*Test.js'`

This has **not** yet been confirmed through the standard `npm run test:container` path.

## Investigation Sketch

- Determine whether the ad hoc full-suite path differs materially from `test/bin/test-in-container.sh`.
- Check whether a failing shell invocation or Mongo client connection can leave MongoDB unavailable for later suites.
[...]

PR #372, opened 2026-05-08
Add container observability: mongod watchdog, EXIT trap, liveness hook, and faster failure messaging (#368)

JamesCropcho:
## Summary

This PR lands the observability and diagnostics layer requested in #368. The root cause of the reported full-suite container timeout cascade was subsequently confirmed as a mongod 8.x SIGSEGV (exit status 139) under the cumulative test workload — the previous harness gave no indication of this, making the crash look like a sprawl of unrelated assertion failures and hook timeouts.

This PR does **not** attempt to mitigate the 8.x mongod segfault itself. That is a workload/version issue that deserves its own dedicated investigation once this observability layer is in place and can provide cleaner reproduction data. See the investigation sketch in #368 for the planned follow-up.

## Changes

**Container init (`docker/init.sh`):**
- Print mongod version, Node.js version, and `uname` at run start so every failure log is self-describing.
- `ulimit -c unlimited` before mongod so a future segfault can leave an inspectable core dump (location is host-controlled via `/proc/sys/kernel/core_pattern`).
- **Mongod watchdog**: background loop monitors `MONGOD_PID` during the test run. When mongod exits, mocha is killed immediately and a clear `FATAL: mongod exited unexpectedly` banner is printed _before_ the cascade starts.
- **EXIT trap**: on any non-zero exit, decodes mongod's exit status including signal name (139 → SIGSEGV), dumps the full mongod log, and prints a memory/process snapshot. All output reaches stdout before `--rm` can remove the container.
- Mocha is now backgrounded (`npm run test:mocha &`) so the watchdog and EXIT trap can act on its lifecycle.

**Container runner (`test/bin/test-in-container.sh`):**
- Drop `--rm`; run cleanup explicitly so `docker cp /tmp/variety-mongod.log ./variety-mongod-debug.log` can run before container removal on failure.
- Pre-remove any leftover same-named container from a prior failed run to avoid name-conflict errors.

**Mocha root hook (`test/helpers/mongod-liveness-hook.cjs`):**
- Pings mongod before every test over a single long-lived `MongoClient`. When the ping fails, the test fails immediately with a clear "MongoDB unreachable on port N — mongod may have crashed" message instead of waiting out the full 15 s mocha timeout.
- If mongod is not reachable at suite start (local dev without a server), the hook silently disables itself — non-container workflows are unaffected.
- Wired via `--require test/helpers/mongod-liveness-hook.cjs` in `test:mocha`. `.eslint.config.js` extended to cover `**/*.cjs` with `globals.node`.

**Test harness (`test/helpers/VarietyHarness.js`):**
- `serverSelectionTimeoutMS` and `connectTimeoutMS` reduced from the 30 s driver default to 3 s. A dead mongod now produces a clear `MongoServerSelectionError` within the test's own error budget rather than a silent mocha timeout.

**Test harness (`test/helpers/MongoShellRunner.js`):**
- Connection-failure patterns (`ECONNREFUSED`, `MongoServerSelectionError`, etc.) in mongosh output are detected and rethrown as a descriptive "mongosh could not connect to MongoDB on port N" error, distinguishing a server-unreachable failure from a variety.js script error. Previously a dead mongod first surfaced as "expected stdout to include `<variety error string>`" — the confusing symptom that #368 originally observed in `ParametersParsingTest`.

**Docs (`CONTRIBUTING.md`):**
- New "Diagnosing container test failures" subsection documents all new diagnostic banners, the `variety-mongod-debug.log` artifact, and the watchdog banner so the next person hitting these symptoms knows where to look immediately.

## Test plan

- [ ] `npm run lint:pre-commit` passes (shell, ESLint, SPDX, markdown, typecheck, build verify)
- [ ] Full container run on MongoDB 8.2 / Node.js 22 produces the `FATAL: mongod exited unexpectedly` banner and mongod exit status in the failure output (rather than a cascade of unrelated timeouts)
- [ ] `npm run test:container` on MongoDB 5.0 / Node.js 22 still passes (5.0 does not exhibit the 8.x crash)
- [ ] `npm run test:mocha` locally (without a running mongod) completes without the liveness hook interfering

Refs #368. The underlying 8.x mongod crash is tracked for a follow-up PR.


JamesCropcho:
Overall: I think this is a sane and reasonably scoped PR. It is doing the right thing for #368 by adding observability/failure attribution without trying to solve the underlying MongoDB 8.x segfault in the same breath. The pieces also feel cohesive to me: the init-script watchdog/trap gives the root cause, the liveness hook and shorter driver timeouts reduce the amount of misleading Mocha noise, the shell-runner message makes connection failures less mysterious, and the CONTRIBUTING update gives future humans/agents somewhere to land.

A few things I would tweak before merging:

- `test/bin/test-in-container.sh` now writes `./variety-mongod-debug.log` on failure, but `.gitignore` does not ignore it. Since this is an expected generated diagnostic artifact, I would add it to `.gitignore` in this PR.
- The watchdog banner says `See '=== mongod exited ===' above`, but the trap prints the status later, and the exact heading is `=== mongod exited with status ... ===`. Tiny thing, but I would change that wording to something like “See the mongod exit diagnostics below” so the log reads correctly during a crash.
- Since dropping `--rm` makes container cleanup explicit, I would consider a small trap in `test/bin/test-in-container.sh` so Ctrl-C or an unexpected script interruption still removes the named container. The normal success/failure path is handled, and the pre-run `rm` helps with stopped leftovers, but interruption is the one lifecycle edge left a bit loose.

One nuance, not necessarily blocking: the new `MongoShellRunner` connection-failure error is a good idea, but tests that intentionally catch runner errors and then assert on `error.code` / `error.stdout` may still turn that clearer error into a local assertion failure if mongod dies mid-test. The watchdog/trap should still make the root cause obvious in the full output, so I do not think this invalidates the approach. I would just avoid overclaiming that this specific assertion shape is fully eliminated unless those catch blocks are also adjusted.

I do not see anything here that should be split out, other than keeping the actual MongoDB crash mitigation/investigation as a follow-up, which the PR already does. The liveness hook is broad, but it disables itself when MongoDB is absent and the per-test ping cost is small, so that tradeoff seems acceptable for this test harness.

JamesCropcho:
Addressed the three concrete proposals in `d912e86`:

- **`.gitignore`**: added `/variety-mongod-debug.log` so the extracted diagnostic log is not accidentally staged.
- **Watchdog banner**: changed "See `=== mongod exited ===` above" → "See mongod exit diagnostics below" — the EXIT trap output follows the banner, so "below" is correct and the exact heading no longer needs to match.
- **Interruption trap in `test-in-container.sh`**: added `trap cleanup_container INT TERM` (defined just before the run) so Ctrl-C or SIGTERM still removes the named container.

On the MongoShellRunner nuance — agreed, the connection-failure rethrow narrows the misleading-output case but does not eliminate it for test catch blocks that assert on `error.code`/`error.stdout`. The watchdog banner is the reliable root-cause signal either way. I've left the code as-is since the reviewer flagged it as informational rather than a requested change, but I'll keep it in mind as a follow-up if those catch blocks ever get adjusted.

JamesCropcho:
Re-reviewed current head through `7eb1fb1`.

This addresses my concrete review points: the extracted diagnostic log is ignored, the watchdog banner now matches the actual order of the diagnostics, and the runner has an interruption cleanup path. I also re-read the surrounding shell flow and the updated comments; I do not see a new blocker or anything that changes my earlier gestalt read. The PR still feels appropriately scoped as diagnostics/observability for #368, with the actual MongoDB 8.x crash left for a separate follow-up.

One tiny non-blocking edge remains around interruption cleanup: `cleanup_container` uses plain `rm`, so if an interrupted attached run somehow leaves the container still running, cleanup may not remove it until the next invocation's pre-run cleanup path. That is not enough for me to ask for another change here, since the normal success/failure path and stopped-leftover path are covered; `rm -f` would just be the belt-and-suspenders version.

No further requested changes from me.

